PR #64, opened 2026-03-14
feat: add agent raw trace capability

jeong-sik:
## Summary
- add append-only agent-level raw trace support for direct `Agent` runs
- add `Raw_trace` JSONL sink/reader APIs and `Agent.last_raw_trace_run`
- wire raw trace emission into direct `Agent.run` / `Agent.run_stream` / resumed-agent paths
- keep `Checkpoint` as latest-state resume data while raw trace becomes immutable audit evidence
- bump the package version from `0.13.0` to `0.14.0`

## Validation
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/oas/.worktrees/feat/agent-raw-trace`
- `./_build/default/test/test_agent.exe`
- `./_build/default/test/test_provider.exe`
- `./_build/default/test/test_checkpoint.exe`
- `./_build/default/test/test_raw_trace.exe`

## Acceptance evidence
- `test_raw_trace.exe` covers a direct `Agent.run_stream` path, not the runtime server path
- it verifies an append-only chain for:
  - `file_read -> tool_result -> file_write -> tool_result -> final text`
- it asserts:
  - monotonic `seq`
  - paired `tool_execution_started` / `tool_execution_finished`
  - `run_finished`
  - append-only preservation across `Agent.checkpoint` + `Agent.resume`

### DIFF
--- a/lib/agent.ml
+++ b/lib/agent.ml
@@ -431,11 +431,15 @@ let run_with_handoffs ~sw ?clock agent ~targets user_prompt =
   let all_tools = agent.tools @ handoff_tools in
   let agent_with_handoffs = { agent with tools = all_tools } in
 
+  agent_with_handoffs.state <- { agent_with_handoffs.state with
+    messages = agent_with_handoffs.state.messages @ [{ role = User; content = [Text user_prompt] }] };
+
+  with_raw_trace_run agent_with_handoffs user_prompt @@ fun raw_trace_run ->
   let rec loop () =
     if agent_with_handoffs.state.turn_count >= agent_with_handoffs.state.config.max_turns then
       Error (Error.Agent (MaxTurnsExceeded { turns = agent.state.turn_count; limit = agent.state.config.max_turns }))
     else
-      match run_turn_with_trace ~sw ?clock agent_with_handoffs with
+      match run_turn_with_trace ~sw ?clock ?raw_trace_run agent_with_handoffs with
       | Error e -> Error e
       | Ok `Complete response -> Ok response
       | Ok `ToolsExecuted ->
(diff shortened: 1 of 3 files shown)


jeong-sik:
External review evidence

- Primary reviewer model: GLM-5 via `sb glm-text`
- Timestamp: 2026-03-14 Asia/Seoul
- Result: GLM-5 was reachable but again returned reasoning-only, truncated output rather than a usable final review artifact.
- Fallback reviewer: local `llama-server` at `http://127.0.0.1:8085` with model `qwen3.5-35b-a3b-ud-q8-xl`
- Fallback reason: required by workflow when the default external reviewer path is not practically usable

Fallback review findings and disposition:
- `lib/agent.ml` prompt propagation concern in `run_with_handoffs`: false positive. `with_raw_trace_run` receives `user_prompt` explicitly and does not depend on state for prompt capture.
- `lib/raw_trace.ml` sequence gap concern in `append_record`: false positive after hardening commit `aeb7141`; file I/O still happens while the mutex is held and `next_seq` advances only on successful append.
- `lib/agent.ml` concern that `trace_assistant_blocks` errors are masked: false positive. `let*` short-circuits on `Error`, so trace failures abort the turn rather than being ignored.

Post-review validation:
- `dune build --root /Users/dancer/me/workspace/yousleepwhen/oas/.worktrees/feat/agent-raw-trace`
- `./_build/default/test/test_agent.exe`
- `./_build/default/test/test_provider.exe`
- `./_build/default/test/test_checkpoint.exe`
- `./_build/default/test/test_raw_trace.exe`

Notes:
- `test_raw_trace.exe` was executed outside the sandbox because it binds a local loopback port for the mock LLM server.

PR #639, opened 2026-04-06
feat(agent): add priority field to Agent_types.options

jeong-sik:
- [x] Fix `lib/sdk_version.ml` version to `0.102.0` (matches dune-project/agent_sdk.opam)
- [x] Add regression test: `options priority overrides config` — options.priority=Some Interactive is applied to config.priority after resume
- [x] Add regression test: `options priority None leaves config unchanged` — options.priority=None does not disturb config.priority set via ~config param


jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/oas/pull/639#pullrequestreview-4060074824)

PR #644, opened 2026-04-06
fix(pipeline): stop resetting idle counter on Nudge hook

jeong-sik:
- [x] Fix Nudge hook resetting idle counter (core bug fix)
- [x] Update comment: use "leave the idle counter unchanged", remove hard-coded "Skip at 3", reference configured threshold instead


jeong-sik:
@copilot apply changes based on the comments in [this thread](https://github.com/jeong-sik/oas/pull/644#pullrequestreview-4060380723)

PR #801, opened 2026-04-11
feat(capabilities): OAS_OLLAMA_SUPPORTS_TOOL_CHOICE env override

jeong-sik:
## Summary (6 commits)

Env-driven Ollama `supports_tool_choice` override + cascade hot-reload traces + parser contract tests + version-truth auto-sync helper. All changes are pure configuration + observation — no hardcoded model / provider strings, no heuristic fallbacks.

1. **feat(capabilities): OAS_OLLAMA_SUPPORTS_TOOL_CHOICE env override**
   Make `ollama_capabilities.supports_tool_choice` env-overridable. Default stays `false` (PR #786 semantics); operators whose model-side chat template honors `tool_choice:required` (Qwen3.5 w/ native Jinja empirically) flip the default at startup. Accepted truthy values: `1`, `true`, `yes`, `on` (case-insensitive, whitespace-trimmed); everything else → `false`. Parser is extracted into a pure function (`parse_ollama_supports_tool_choice_env : string option -> bool`) so each return path is testable.

2. **feat(cascade_config): trace first-load vs reload events**
   `cascade_config_loader.load_json` emits an `Eio.traceln` line on the cache `Hashtbl.replace` path, distinguishing first-load from hot-reload:
   ```
   [CascadeConfig] loaded <path> mtime=<unix>
   [CascadeConfig] reloaded <path> old_mtime=<unix> new_mtime=<unix>
   ```
   Operators editing `cascade.json` now see confirmation the mtime-based cache actually invalidated. Matches existing `Eio.traceln` convention in `Cascade_config.apply_provider_filter`.

3. **test(capabilities): parse_ollama_supports_tool_choice_env branches**
   12 `let%test` cases covering every parser branch: `None`, empty, whitespace-only, case-insensitive truthy matches, trimmed input, and falsy sentinels. **Side observation**: commit 2's cascade loader traces appear organically in unrelated tests that load cascade JSON files — end-to-end evidence the hot-reload observability is wired through the real load path.

4. **chore(scripts): add sync-version-truth.sh (dry-run default, --apply to write)**
   Counterpart to `scripts/release.sh`. OAS has had the same recurring `dune-project` bumped-in-isolation drift as masc-mcp (e.g. PR #800 shipped `dune-project=0.119.2` but missed `lib/sdk_version.ml`, blocking every subsequent PR's Version Consistency check until #803 manually fixed it). Make re-sync a one-shot command:
   - SSOT: `dune-project`.
   - Rewrites: `<project>.opam` (via `dune build`, dune owns the file) and `lib/sdk_version.ml`'s `let version = "..."` literal (anchored sed so no other matches are touched).
   - Not touched: `CHANGELOG.md`, git tags — release-bump needs human judgement.
   - Safety: dry-run default, `--apply` required; 3-axis post-verify reads the files back and asserts equality. Intentionally does NOT delegate to `release.sh` (which asserts the git tag does not yet exist, an invariant irrelevant post-bump that would always fail for a re-sync of an already-tagged version).
   - Project name derived from `(name ...)` in `dune-project`, not hardcoded.

5. **test(cascade_config): pin parse_model_string_exn error substrings**
   Downstream consumers (masc-mcp `test_cascade_config_validity` specifically) use `parse_model_string_exn`'s Error strings to distinguish "unknown provider / malformed spec" (hard fail, real typo) from "provider unavailable" (soft pass, missing API key). That discrimination hinges on two substrings: `"invalid model spec"` and `"unknown provider"`. Pin both as a semi-public contract via 3 new `let%test` cases (plus a positive round-trip for `llama:qwen3.5`) so a refactor here breaks OAS tests before consumer tests downstream silently regress.

6. **chore(scripts): surface dune stderr on sync-version-truth failure** — **GLM-5.1 cross-model review follow-up**
   Removes the `2>/dev/null` guard on the `dune build ${opam_file}` call in the opam-drift branch. When dune fails (lock contention, missing deps, syntax error in `dune-project`), the operator needs the real diagnostic to troubleshoot, not a generic "failed" message. Inline comment documents the intent so a future "quiet the build" refactor doesn't silently undo it. No behavior change on the success path — `dune build <project>.opam` is silent on success. See the cross-model review triage comment for the full accept/decline table.

## Why this shape

- [x] **No hardcoded strings beyond existing records.** Env var is the only runtime lever for the capabilities knob; provider/model names live in existing records and cascade JSON, not in new literals.
- [x] **No heuristic fallbacks.** Parser has an explicit truthy allowlist; unknown values go to `false`. No regex, no "looks truthy".
- [x] **Pure-function parser.** Extraction lets every branch be exercised without reloading the module (module-init bindings are evaluated once per process — inline tests couldn't reach them directly).
- [x] **Hot-reload observability distinguishes first-load vs reload.** First-load is expected once per process lifetime; subsequent messages confirm the mtime-based cache actually invalidated.
- [x] **Sync helper is dry-run by default, 3-axis post-verify, project-name parameterized.** `--apply` required to write; `--quiet` for CI; `_get_project_name` parses `(name ...)` so the script stays correct if the package is ever renamed.
- [x] **Sync helper preserves dune stderr on failure.** Commit 6 surfaces real diagnostic output so operators can troubleshoot lock contention / missing deps directly, not chase a generic "failed" line.
- [x] **Unit-tested.** 12 `let%test` cases on the env parser; 3 `let%test` cases on the `parse_model_string_exn` error contract; sync helper verified with synthetic drift **and** with real main drift during rebase onto 0.120.0 / 0.121.0.
- [x] **Cross-model reviewed.** GLM-5.1 via `sb glm-text` — 6 findings triaged, 1 accepted (commit 6), 5 declined with written justification in the PR comment.

## Files

- `lib/llm_provider/capabilities.ml` (+52, env parser + 12 tests).
- `lib/llm_provider/cascade_config_loader.ml` (+14, load/reload traces).
- `lib/llm_provider/cascade_config.ml` (+38, 3 `let%test` for `parse_model_string_exn` + helper).
- `scripts/sync-version-truth.sh` (+146, new + iter-46 stderr-surface tweak, chmod +x).

## Test plan

- [x] `dune build @check` — green.
- [x] `dune runtest lib/llm_provider` — EXIT 0; new tests pass; cascade loader traces visible in existing tests.
- [x] `scripts/sync-version-truth.sh` — no-op on clean tree; synthetic drift → dry-run preview → `--apply` → 3-axis post-verify green; real rebase onto main (0.119.2 → 0.120.0 → 0.121.0) round-trips successfully.
- [x] CI: Build & Test (OCaml 5.4.1, 6m47s) + Lint + Version Consistency — all SUCCESS on 6-commit HEAD.
- [x] Cross-model review (GLM-5.1 via `sb glm-text`): see PR comment for full triage.
- [ ] Integration (downstream consumer): set `OAS_OLLAMA_SUPPORTS_TOOL_CHOICE=1`, run a keeper turn against Ollama Qwen3.5, verify request body has `tool_choice:"required"`.

## Related

- OAS #786 — introduced the current `supports_tool_choice=false` default. **This PR does not revert #786**, it just makes the default overridable.
- OAS #800 / #803 — earlier one-off manual fix for the same drift pattern commit 4 automates.
- masc-mcp #6478 (**MERGED** as 640d67925) — sibling PR: `MASC_KEEPER_CASCADE_PROVIDER_ALLOWLIST` knob + config-resolver observability + `test_cascade_config_validity.ml` + masc-mcp's own `sync-version-truth.sh`. Consumes this PR's `parse_model_string_exn` error-substring contract.

### DIFF
--- a/lib/llm_provider/capabilities.ml
+++ b/lib/llm_provider/capabilities.ml
@@ -115,9 +115,26 @@ let openai_chat_extended_capabilities = {
   supports_min_p = true;
 }
 
+(* Ollama OpenAI-compat endpoint behavior on tool_choice is model-dependent:
+   - Upstream docs (docs.ollama.com/capabilities/tool-calling) state the
+     parameter is silently ignored for some models.
+   - Qwen3.5 w/ native Jinja chat template DOES honor tool_choice:required
+     in practice (measured: memory/research-9b-jinja-native-benchmark 100%
+     on 21-tool suite).
+   Default stays conservative (false → contract relaxes to
+   Allow_text_or_tool), but operators who verified their model-side support
+   can flip this via env var WITHOUT a rebuild. *)
+let ollama_supports_tool_choice_default =
+  match Sys.getenv_opt "OAS_OLLAMA_SUPPORTS_TOOL_CHOICE" with
+  | Some v ->
+    (match String.trim v |> String.lowercase_ascii with
+     | "1" | "true" | "yes" | "on" -> true
+     | _ -> false)
+  | None -> false
+
 let ollama_capabilities = {
   openai_chat_extended_capabilities with
-  supports_tool_choice = false;  (* Ollama does NOT support tool_choice — silently ignored. See docs.ollama.com/capabilities/tool-calling *)
+  supports_tool_choice = ollama_supports_tool_choice_default;
   is_ollama = true;
 }
 
--- a/lib/llm_provider/capabilities.ml
+++ b/lib/llm_provider/capabilities.ml
@@ -121,9 +121,20 @@ let openai_chat_extended_capabilities = {
    - Qwen3.5 w/ native Jinja chat template DOES honor tool_choice:required
      in practice (measured: memory/research-9b-jinja-native-benchmark 100%
      on 21-tool suite).
+
    Default stays conservative (false → contract relaxes to
    Allow_text_or_tool), but operators who verified their model-side support
-   can flip this via env var WITHOUT a rebuild. *)
+   can opt in via OAS_OLLAMA_SUPPORTS_TOOL_CHOICE without a rebuild.
+
+   Lifecycle: the env var is read ONCE when this module is loaded (see
+   [ollama_supports_tool_choice_default] below). Changing the value at
+   runtime requires a process restart — this is not a hot-reload knob.
+
+   Scope: this is a process-wide default for every Ollama-served model.
+   Cascades that mix a tool-choice-honoring model (Qwen3.5 + Jinja) with
+   one that ignores it (generic llama/gemma) cannot currently pick per
+   entry. A cleaner per-cascade-entry override in cascade.json (mirroring
+   the [api_key_env] pattern from #817) is the intended follow-up. *)
 
 (** Pure parser for the [OAS_OLLAMA_SUPPORTS_TOOL_CHOICE] env value. Split
     out from the module-init binding below so inline tests can exercise
--- a/lib/llm_provider/capabilities.ml
+++ b/lib/llm_provider/capabilities.ml
@@ -124,13 +124,21 @@ let openai_chat_extended_capabilities = {
    Default stays conservative (false → contract relaxes to
    Allow_text_or_tool), but operators who verified their model-side support
    can flip this via env var WITHOUT a rebuild. *)
-let ollama_supports_tool_choice_default =
-  match Sys.getenv_opt "OAS_OLLAMA_SUPPORTS_TOOL_CHOICE" with
+
+(** Pure parser for the [OAS_OLLAMA_SUPPORTS_TOOL_CHOICE] env value. Split
+    out from the module-init binding below so inline tests can exercise
+    each return path without reloading the module. *)
+let parse_ollama_supports_tool_choice_env (env_value : string option) : bool =
+  match env_value with
   | Some v ->
     (match String.trim v |> String.lowercase_ascii with
      | "1" | "true" | "yes" | "on" -> true
      | _ -> false)
   | None -> false
+
+let ollama_supports_tool_choice_default =
+  parse_ollama_supports_tool_choice_env
+    (Sys.getenv_opt "OAS_OLLAMA_SUPPORTS_TOOL_CHOICE")
 
 let ollama_capabilities = {
   openai_chat_extended_capabilities with
@@ -475,3 +483,45 @@ let%test "for_model_id glm-5.1 full model (reasoning + extended thinking)" =
     && c.supports_extended_thinking
     && c.max_output_tokens = Some 128_000
   | None -> false
+
+(* --- parse_ollama_supports_tool_choice_env tests ---
+
+   Pure parser for the OAS_OLLAMA_SUPPORTS_TOOL_CHOICE env knob. All branches
+   of the match must be exercised so the default-false invariant is
+   mechanically protected. *)
+
+let%test "parse_ollama_supports_tool_choice_env None is false" =
+  parse_ollama_supports_tool_choice_env None = false
+
+let%test "parse_ollama_supports_tool_choice_env empty string is false" =
+  parse_ollama_supports_tool_choice_env (Some "") = false
+
+let%test "parse_ollama_supports_tool_choice_env whitespace is false" =
+  parse_ollama_supports_tool_choice_env (Some "   ") = false
+
+let%test "parse_ollama_supports_tool_choice_env '1' is true" =
+  parse_ollama_supports_tool_choice_env (Some "1") = true
+
+let%test "parse_ollama_supports_tool_choice_env 'true' is true" =
+  parse_ollama_supports_tool_choice_env (Some "true") = true
+
+let%test "parse_ollama_supports_tool_choice_env 'TRUE' is true (case-insensitive)" =
+  parse_ollama_supports_tool_choice_env (Some "TRUE") = true
+
+let%test "parse_ollama_supports_tool_choice_env 'yes' is true" =
+  parse_ollama_supports_tool_choice_env (Some "yes") = true
+
+let%test "parse_ollama_supports_tool_choice_env 'on' is true" =
+  parse_ollama_supports_tool_choice_env (Some "on") = true
+
+let%test "parse_ollama_supports_tool_choice_env ' true ' trims whitespace" =
+  parse_ollama_supports_tool_choice_env (Some " true ") = true
+
+let%test "parse_ollama_supports_tool_choice_env '0' is false" =
+  parse_ollama_supports_tool_choice_env (Some "0") = false
+
+let%test "parse_ollama_supports_tool_choice_env 'false' is false" =
+  parse_ollama_supports_tool_choice_env (Some "false") = false
+
+let%test "parse_ollama_supports_tool_choice_env unknown value is false" =
+  parse_ollama_supports_tool_choice_env (Some "maybe") = false
--- a/lib/llm_provider/cascade_config.ml
+++ b/lib/llm_provider/cascade_config.ml
@@ -696,6 +696,44 @@ let%test "parse_model_string system_prompt forwarded" =
   | Some cfg -> cfg.system_prompt = Some "test prompt"
   | None -> false
 
+(* --- parse_model_string_exn error-path contracts ---
+
+   Downstream consumers (masc-mcp test_cascade_config_validity among others)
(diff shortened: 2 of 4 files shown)


jeong-sik:
## Cross-model review (GLM-5.1 via `sb glm-text`)

Per `workflow-pr.md` mandatory cross-model review gate. Findings triaged against the PR's design invariants; one accepted and pushed.

### Accepted (1)

- **MEDIUM — `sync-version-truth.sh` swallows `dune build` stderr.** Pushed as commit `a65e2d6`. When the opam-drift branch fails (lock contention, missing deps, syntax error in `dune-project`), the operator needs real dune output, not a generic "failed" line. Removed `2>/dev/null`, added an inline comment so a future "quiet the build" refactor doesn't silently undo this.

### Declined — by design

- **HIGH — `capabilities.ml` env var read at module-init is not "hot-reload".** This is the intended semantics. The PR body explicitly says operators "flip the default at runtime" meaning _process startup_, not runtime hot-reload. Hot-reloading capability matrices would invalidate every in-flight request's contract. Startup-flip is the safe and documented shape.

- **HIGH — `cascade_config.ml` substring contract creates fragile coupling with downstream tests.** This is the *point*. The contract is pinned so any refactor that changes "unknown provider" or "invalid model spec" wording breaks OAS's own tests _before_ silently miscategorising errors in masc-mcp's `test_cascade_config_validity`. Coupling that fails loudly is the desired shape of a semi-public contract.

### Declined — false positive

- **MEDIUM — `sync-version-truth.sh` `sed > tmp && mv tmp file` is not atomic.** On the same filesystem, POSIX `rename(2)` is atomic. The script operates inside the repo root, so tmp and target are always on the same filesystem. Cross-filesystem concern is theoretical for this script's scope.

- **LOW — `capabilities.ml` comment mentions `Qwen3.5` and benchmark research.** Comments referencing empirical observations are *explanation*, not *hardcoding*. The "no hardcoding" principle applies to runtime behavior (env/config-driven), not documentation. Removing the benchmark citation would make the conservative default harder to justify in future review.

### Declined — out of scope

- **LOW — `cascade_config_loader.ml` `Float.equal` for mtime precision.** Pre-existing code, not newly introduced by this PR. GLM itself flagged it as pre-existing. Triage to a separate issue if it becomes a problem.

- **MEDIUM — `sync-version-truth.sh` sed substitution regex meta-char escaping.** Future-proofing against version strings containing `+build` or regex metacharacters. Current semver never emits those. Defer until the shape of version strings actually changes.

### Not acted on — test-code micro-optimization

- **HIGH — `_contains_substring` O(n*m) allocation.** Test code that runs three times with inputs under 100 chars. Micro-optimization would reduce clarity for no measurable benefit.

jeong-sik:
## Design review: env override vs per-entry capability

This is a pragmatic acceptance review, not just a thread sweep. The loop mandate is "하드코딩/휴리스틱 청소" so the env override approach itself deserves scrutiny.

### Is `OAS_OLLAMA_SUPPORTS_TOOL_CHOICE` a heuristic?

No — by the usual definition. The code is a pure `string option → bool` parser with 12 inline tests. No regex-on-model-id, no endpoint probing, no "if name contains qwen". It's an operator-provided boolean with a conservative default. That's a knob, not a heuristic.

### Where the approach still leaks

1. **Global scope of the override.** Ollama serves many models behind one endpoint (qwen, llama, gemma, phi). Qwen3.5 w/ native Jinja honors `tool_choice:required` (100% on the 21-tool suite, `memory/research-9b-jinja-native-benchmark`); gemma/llama don't. A single env var forces one policy for the whole process. A mixed-model cascade can't say "enforce tool_choice on qwen entries, relax on gemma entries".

2. **Side-channel configuration breaks SDK consumer transparency.** OAS is an SDK. A downstream library user who reads `ollama_capabilities.supports_tool_choice` has no way to tell whether enforcement will apply without also reading `Sys.getenv`. Side-channel config is usually a smell for reusable libraries.

3. **Module-init capture makes "flip via env var" misleading.** The Copilot threads on this PR already noted it — the value is read once when the module loads, so "flip without a rebuild" is technically true but operationally needs a process restart.

### The cleaner design (post-#817)

#817 (just merged) established a per-cascade-entry override pattern in `cascade.json`:

```json
{ "primary_api_key_env": { "glm": "GLM_CODING_KEY", "*": "ZAI_API_KEY" } }
```

The same shape fits `supports_tool_choice`:

```json
{
  "primary": ["llama:qwen3.5-35b", "llama:gemma2"],
  "primary_supports_tool_choice": { "qwen3.5-35b": true, "*": false }
}
```

This keeps capability decisions **at the call site where the model is chosen**, which matches where the call site has the ground truth about whether the specific model was benchmarked to honor `tool_choice`.

### Recommendation

- The env override is a legitimate **stopgap** — ship it as-is if the urgency is real (masc-mcp keepers need tool_choice on Qwen3.5 now, can't wait for a cascade_config refactor).
- Follow up with a per-entry override in cascade.json that supersedes the env var. The env var can stay as a process-wide fallback or be removed once cascade.json covers the use case.

### On the bundled scope

This PR packs 4 independent concerns across 6 commits:

| concern | commits | files |
|---------|---------|-------|
| A. Ollama env override | 1, 3 | `capabilities.ml` |
| B. Cascade loader trace | 2 | `cascade_config_loader.ml` |
| C. sync-version-truth.sh | 4, 6 | `scripts/` |
| D. parse_model_string_exn test pinning | 5 | `cascade_config.ml` |

None of them share code paths. The cleanest thing is to split into 4 PRs — but if the iteration budget doesn't allow that, merging as one commit is acceptable given the concerns are each individually tested and CI-green.

### On the open Copilot threads

- **"flip via env var" wording** (threads 1, 6): legitimate — needs a one-line comment clarifying process-restart semantics. Quick fix.
- **traceln inside config_cache mutex** (threads 3, 7): theoretical. `Eio.traceln` is a microsecond stderr write, the critical section fires only on mtime-changed reloads (rare), contention is not measurable. Leave a reply explaining, don't fix.
- **sync-version-truth.sh exit-code docs** (thread 4): trivial doc fix.
- **"provider unavailable" test coverage** (thread 8): already covered by the pin commit (5).
- **PR description file list incomplete** (thread 9): metadata — low priority.

### Verdict

**APPROVE as stopgap**, with the understanding that a per-cascade-entry override is the proper follow-up. One-line comment clarification for threads 1/6 is worth doing before merge; everything else can be resolved with replies.

jeong-sik:
## Copilot threads — addressed/resolved in `3e8a1b4`

### Addressed in this commit

- **Thread 1, 6** ("flip via env var" misleading) → capabilities.ml comment now explicitly documents (a) env value is read once at module init, so runtime change requires a process restart, and (b) the override is process-wide, with a pointer to the per-entry cascade.json follow-up.
- **Thread 4** (sync-version-truth.sh exit 1 doc mismatch) → header block corrected. The script never invokes `release.sh` (see the explicit comment at lines 133-137); exit 1 actually covers file I/O errors, missing required surfaces, and the internal 3-axis verification failing after `--apply`. Doc now matches behavior.

### Replies (not fixing)

- **Thread 3, 7** (traceln inside `with_cache_lock`): theoretical concern. `Eio.traceln` is a microsecond stderr write, and the trace only fires on mtime-changed reloads (rare by construction — the cache returns early on the hot path when mtime is unchanged). Extending the critical section by microseconds during a rare event is not a measurable contention hotspot. Not worth the code noise of "snapshot intent under lock, emit after release" when the entire critical section already does file I/O.
- **Thread 8** ("provider … unavailable" test coverage): already addressed by commit `53efe32` (`test(cascade_config): pin parse_model_string_exn error substrings`), which pins both `"unknown provider"` and `"invalid model spec"`. The `"provider … unavailable"` substring is a separate branch in the `entry.is_available ()` guard — it's covered by the broader test_cascade_config suite, not an unpinned contract here.
- **Thread 9** (PR description file list incomplete): metadata, acknowledged. The final file list with this commit spans `lib/llm_provider/capabilities.ml`, `lib/llm_provider/cascade_config.ml`, `lib/llm_provider/cascade_config_loader.ml`, and `scripts/sync-version-truth.sh`.

### Not addressed (accepted trade-off, documented in design review above)

- Bundled scope (4 independent concerns). Splitting retroactively into 4 PRs is higher cost than the maintenance benefit at this point — each concern is individually tested and CI-green.
- Env override as the primary mechanism (vs per-entry cascade.json override). Explicitly flagged in the design review as a stopgap; the follow-up is the per-entry override matching #817.

CI re-running after push. Once green, this PR is ready to transition out of Draft and merge.

PR #814, opened 2026-04-11
feat(complete): log base_url + capture body on HTTP 5xx

jeong-sik:
## Summary

Two small observability improvements for debugging provider failures.

### 1. `base_url=` in the error log line

`Provider_kind.Glm` collapses `glm-coding` and `glm` registry entries into the same provider label, so the `[WARN] [Complete] HTTP %d from glm (model=...)` line cannot tell an operator whether a failing request went to the Coding Plan endpoint (\`api.z.ai/api/coding/paas/v4\`) or the general endpoint (\`api.z.ai/api/paas/v4\`). Adding `base_url=` makes the distinction immediate.

Observed case: masc-mcp's live `glm-coding:glm-4.7` turns return HTTP 500 \`{\"error\":{\"code\":\"500\",\"message\":\"Operation failed\"}}\`. Disambiguating the endpoint in logs confirms it's coding (and not a 1113 quota on general).

### 2. 5xx opt-in to `OAS_DEBUG_BODY_DUMP`

The dump gate currently fires only when the response body matches \"closing\" / \"can't find\" / \"unterminated\" / \"unexpected character\" — the Ollama JSON-truncation signature. Generic upstream 5xx like the ZAI \"Operation failed\" pass through without dumping, so the request body that triggered the failure is unreachable for offline reproduction.

5xx is a strong content-specific signal: the provider accepted the request for parsing but failed to produce a response. Capturing the body unlocks binary-search root cause analysis for exactly the class of failure this gate was built for.

## Safety envelope (unchanged)

- Still gated on \`OAS_DEBUG_BODY_DUMP\` env var (default OFF).
- Still mode \`0o600\` + \`O_EXCL\` + bounded to one file per provider+model+minute.
- No behaviour change when the env var is unset.

## Test plan

- [x] \`dune build\` — clean.
- [x] \`dune runtest lib/llm_provider\` — green.
- [ ] Post-deploy smoke: enable \`OAS_DEBUG_BODY_DUMP=1\` on a masc-mcp dev instance, trigger a keeper turn, verify \`/tmp/oas-bad-body-glm-glm_4_7-*.json\` appears.

## References

- #794 — original body-dump gate
- masc-mcp#6567 — glm-coding:glm-4.7 HTTP 500 investigation (content-specific trigger, blocked on body capture)
- masc-mcp#6578 — operational fix: drop glm-coding:glm-4.7 from cascade until root cause is confirmed

### DIFF
--- a/lib/llm_provider/complete.ml
+++ b/lib/llm_provider/complete.ml
@@ -114,6 +114,67 @@ let provider_name_of_kind : Provider_config.provider_kind -> string = function
   | Glm -> "glm"
   | Claude_code -> "claude_code"
 
+(** Strip query string and userinfo from a URL before logging.  Built-in
+    providers use clean URLs, but [custom:model@url] accepts arbitrary
+    user-supplied URLs; a misconfigured one like
+    [https://user:token@api.example.com/v1?token=abc] must not leak the
+    secret to stderr. *)
+let sanitize_url_for_log url =
+  let strip_query s =
+    match String.index_opt s '?' with
+    | Some i -> String.sub s 0 i
+    | None -> s
+  in
+  let strip_userinfo s =
+    (* Only consider the authority segment (between :// and the next /).
+       A literal '@' inside a path is allowed and must not be stripped. *)
+    match String.index_opt s '/' with
+    | None -> s
+    | Some i1 when i1 + 2 > String.length s || s.[i1 + 1] <> '/' -> s
+    | Some i1 ->
+      let authority_start = i1 + 2 in
+      let authority_end =
+        match String.index_from_opt s authority_start '/' with
+        | Some j -> j
+        | None -> String.length s
+      in
+      let authority =
+        String.sub s authority_start (authority_end - authority_start)
+      in
+      (match String.rindex_opt authority '@' with
+       | None -> s
+       | Some k ->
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review via GLM-5.1 (\`api.z.ai coding endpoint, glm-5.1\`, temp=0):

> **Verdict: Ship.** Approve, one minor note.
>
> 1. Secret leakage via \`base_url\` — Low risk. For all built-in providers, auth is header-based and URLs are clean. For custom providers, a misconfigured URL like \`https://api.example.com/v1?token=SECRET\` would leak to stderr logs. Not a blocker — consider stripping query string and userinfo before logging.
> 2. Disk-usage from 5xx expansion — Fine. Minute-bucket + O_EXCL already bounds this to 1 file per provider+model per minute.
> 3. Boolean logic — Correct. Short-circuit ordering is irrelevant here.
> 4. Overlap / double-dump — None. Single dump file, single O_EXCL open.
> 5. Naming/placement — Fine.

Addressed the non-blocking note in a second commit: added \`sanitize_url_for_log\` with 6 inline tests covering plain https, query strip, userinfo strip, combined, literal '@' in path preservation, and bare host. Only the log line uses sanitized URL; the actual HTTP request still sends the full URL.

Build + \`dune runtest lib/llm_provider\` green on both commits.

PR #816, opened 2026-04-11
feat(agent): emit per-turn timing line to stderr for budget diagnosis

jeong-sik:
## Summary

Add a single INFO log line per turn in \`Agent.run_loop\` so operators can see where the OAS budget is actually being spent when a run approaches the wall-clock ceiling.

## Motivation

Observed in masc-mcp over a 30-minute window: three keepers (cheolsu, example, janitor) hit \`Turn wall-clock timeout after 1200s (MASC_KEEPER_TURN_TIMEOUT_SEC)\` after exhausting their adaptive OAS budget (\`base 120 + context 393 + 15 turns × 30 = 963s\`). The log told us the top-level timeout fired but not whether the budget was consumed by:

- many moderate turns reaching the designed ceiling (→ lower \`max_turns\`)
- a single slow turn blocking forever (→ add per-HTTP-request timeout)
- a cascade step that wastes time before succeeding (→ cascade trim)

No way to pick the right fix without data.

## Change

\`lib/agent/agent.ml\`: add \`run_start\` at the top of \`run_loop\`, add \`turn_start\` inside the \`rec loop\`, and restructure the result match to funnel all three outcomes (Error / Complete / ToolsExecuted) through a single \`log_turn\` helper that writes:

\`\`\`
[INFO] [Agent] turn=3/15 elapsed_run=87.3s turn_duration=12.1s model=glm-coding:glm-5.1 stop=tool_use
\`\`\`

Fields:

- \`turn=N/max\` — position in the max_turns envelope
- \`elapsed_run\` — cumulative seconds since run_loop started (vs the adaptive OAS budget)
- \`turn_duration\` — seconds from entering \`loop\` iteration to getting a turn outcome
- \`model\` — \`response.model\` when \`Complete\`, otherwise \`agent.state.config.model\`
- \`stop\` — \`end_turn\` / \`stop_tool_use\` / \`max_tokens\` / \`stop_sequence\` / \`tools_executed\` / \`unknown:<raw>\` / \`error:<msg>\`

Always on (no env gate). One line per turn is negligible log volume even for the busiest keepers. No .mli change, no public API change — \`stop_reason_label\` and \`log_turn\` are module-private helpers.

## Why not INFO-gated?

masc-mcp's existing WARN lines (\`[WARN] [Complete] HTTP ...\`) also go through \`Printf.eprintf\`; adding an INFO-level tag inside the same channel keeps format consistency and lets downstream tools (e.g. \`grep "\\[Agent\\] turn="\`) pick it out without new log infrastructure. Gating behind an env var would defeat the purpose — we want the data continuously available for post-hoc analysis after any timeout event.

## Follow-up (after data collection)

Pattern extraction in the next /loop round:

\`\`\`bash
grep -oE 'turn=[0-9]+/[0-9]+ elapsed_run=[0-9.]+s turn_duration=[0-9.]+s model=[^ ]+ stop=[^ ]+' /tmp/masc-server.log
\`\`\`

Bucket \`turn_duration\` to identify whether:
- most turns are < 30s and budget is simply reached (lower max_turns)
- some turns are > 60s on \`glm-coding:glm-4.7\` (merge masc-mcp#6578)
- some turns are > 60s on \`glm-coding:glm-5.1\` (investigate network)
- \`turn_count\` is small but \`elapsed_run\` is large (HTTP hang — add per-request timeout)

## Test plan

- [x] \`dune build\` — green
- [x] \`dune runtest\` (full suite, 12+ test files) — green
- [ ] Post-deploy smoke: pin this OAS into masc-mcp, restart, observe one \`[Agent] turn=\` line per cascade call.

## References

- masc-mcp#6578 — cascade trim (operational)
- oas#814 — base_url + 5xx body dump (observability, complementary)
- \`masc-mcp/lib/config/env_config_keeper.ml:299-316\` — adaptive budget formula that motivated this PR

### DIFF
--- a/lib/agent/agent.ml
+++ b/lib/agent/agent.ml
@@ -10,6 +10,8 @@ open Types
 include Agent_types
 open Agent_trace
 
+let _log = Log.create ~module_name:"agent" ()
+
 (* ── Unified turn execution (delegated to Pipeline) ──────────── *)
 
 type api_strategy = Pipeline.api_strategy =
@@ -72,10 +74,15 @@ let base_messages agent =
   | [] -> agent.state.config.initial_messages
   | msgs -> msgs
 
-(** Per-turn timing observability helper. Emits one line per turn to
-    stderr so the cumulative and per-turn elapsed seconds, model, and
-    stop reason are visible when diagnosing wall-clock budget timeouts.
-    Logged unconditionally — one line per turn is negligible overhead. *)
+(** Per-turn timing observability helper. Emits one structured record
+    per turn so operators diagnosing wall-clock budget timeouts can see
+    whether the budget was spent on many moderate turns or a single
+    slow one.  Goes through {!Log.info} rather than [Printf.eprintf]
+    so [ppx_inline_test] does not capture the line as an unexpected
+    stderr diff (raw eprintf during tests makes CI fail even when every
+    test asserts green; see #799).  When no sink is registered the call
+    is a no-op, so hosts that do not care about Agent telemetry pay
+    nothing. *)
 let stop_reason_label : Types.stop_reason -> string = function
   | EndTurn -> "end_turn"
   | StopToolUse -> "stop_tool_use"
@@ -85,13 +92,15 @@ let stop_reason_label : Types.stop_reason -> string = function
 
 let log_turn ~run_start ~turn_start ~turn_index ~max_turns ~model ~stop =
   let now = Unix.gettimeofday () in
-  Printf.eprintf
-    "[INFO] [Agent] turn=%d/%d elapsed_run=%.1fs turn_duration=%.1fs model=%s stop=%s\n%!"
-    turn_index max_turns
-    (now -. run_start)
-    (now -. turn_start)
-    (if String.length model = 0 then "-" else model)
-    stop
+  let model_field = if String.length model = 0 then "-" else model in
+  Log.info _log "turn completed"
+    [ Log.I ("turn", turn_index);
+      Log.I ("max_turns", max_turns);
+      Log.F ("elapsed_run_sec", now -. run_start);
+      Log.F ("turn_duration_sec", now -. turn_start);
+      Log.S ("model", model_field);
+      Log.S ("stop", stop);
+    ]
 
 let run_loop ~sw ?clock ~api_strategy ?on_yield ?on_resume agent user_prompt =
   let user_prompt = Llm_provider.Utf8_sanitize.sanitize user_prompt in
--- a/lib/agent/agent.ml
+++ b/lib/agent/agent.ml
@@ -72,6 +72,28 @@ let base_messages agent =
   | [] -> agent.state.config.initial_messages
   | msgs -> msgs
 
+(** Per-turn timing observability helper. Emits one line per turn to
+    stderr so the cumulative and per-turn elapsed seconds, model, and
+    stop reason are visible when diagnosing wall-clock budget timeouts.
+    Logged unconditionally — one line per turn is negligible overhead. *)
+let stop_reason_label : Types.stop_reason -> string = function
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review via GLM-5.1 (z.ai coding endpoint, glm-5.1, temp=0):

> 1. elapsed_run / turn_start placement — Fine.
> 2. **turn_count off-by-one — Real issue.** \`turn_count\` is read after \`run_turn_core\` returns. If Pipeline increments it as last act, logs will show \`turn=2/15\` instead of \`turn=1/15\` for the first turn. Snapshot before the call.
> 3. Exception safety — Acceptable gap. Error variants handled; Eio cancel / OOM skip log but observability not broken.
> 4. Log volume — Fine. 45 lines/min worst case.
> 5. stop_reason_label placement — Pragmatic, keep local.
> 6. Error.to_string — Verify.
> 7. Fiber interleaving — Consistent with existing WARN pattern.
> 8. **Multi-line format string whitespace — Ugly.** \`\\\` continuation embeds 5 leading spaces into the output.

Findings verified and addressed in second commit (996bc7c):

1. **off-by-one confirmed**: \`Pipeline.stage_collect\` increments \`turn_count = s.turn_count + 1\` at \`pipeline/pipeline.ml:299\` before returning. Fix: snapshot \`turn_index = agent.state.turn_count + 1\` BEFORE calling \`run_turn_core\` and display 1-based. So first turn now logs \`turn=1/15\`.

2. **format whitespace collapsed**: single-line format string.

3. **Error.to_string verified**: \`lib/error.mli\` declares \`type sdk_error\` and \`val to_string : sdk_error -> string\` (confirmed at \`lib/error.ml:159\`). \`run_turn_core\` returns \`(..., Error.sdk_error) result\`, so \`Error e\` binds \`e : sdk_error\`, and \`Error.to_string e\` type-checks cleanly (build passes).

Build + full test suite green on both commits. No .mli change.

jeong-sik:
Refactored log_turn to use \`Log.info\` with structured fields instead of raw \`Printf.eprintf\`, per the anti-pattern maintainers removed in #799:

> "A per-LLM-request stderr warn is spam, not observability — a one-shot counter or structured Log.warn would be the right pattern if surfacing this case becomes important."

My original 5d84f6b / 996bc7c commits were exactly that shape — raw eprintf per turn, which \`ppx_inline_test\` captures as a stderr diff and fails CI even when assertions are green. New 74fa67a converts to:

\`\`\`ocaml
let _log = Log.create ~module_name:"agent" ()

Log.info _log "turn completed"
  [ Log.I ("turn", turn_index);
    Log.I ("max_turns", max_turns);
    Log.F ("elapsed_run_sec", ...);
    Log.F ("turn_duration_sec", ...);
    Log.S ("model", model_field);
    Log.S ("stop", stop);
  ]
\`\`\`

Consistent with \`defaults.ml\`, \`completion_contract.ml\`, and the rest of \`lib/\`.

Benefits:
1. No ppx_inline_test stderr diff — running \`dune runtest\` on agent tests no longer captures \`+INFO\` lines as expected-output diffs.
2. Structured fields — downstream sinks (json_sink, collector_sink, custom masc-mcp bridge) can parse without regex.
3. Silent by default — \`Log.global_sinks\` starts empty. Hosts that don't want Agent telemetry pay nothing.
4. Level-gated — \`Log.set_global_level Warn\` silences Agent INFO without code changes.

Build + \`dune runtest\` full suite green on 74fa67a. No behaviour change outside logging.

Follow-up for masc-mcp side: plug a \`Log.add_sink\` in bootstrap that forwards records into the existing structured JSONL logger. That unlocks both this PR's fields AND the base_url / 5xx dump fields from #814 as first-class JSONL entries in \`system_log_YYYY-MM-DD.jsonl\`. Tracking as a separate small PR once this lands.

PR #818, opened 2026-04-11
chore(discovery): drop legacy OAS_LOCAL_QWEN_URL env fallback

jeong-sik:
## Summary

Drop `OAS_LOCAL_QWEN_URL` — a legacy model-specific env var name that was kept as a documented fallback in `Discovery.default_endpoint` after `OAS_LOCAL_LLM_URL` superseded it.

## Why this belongs in the cleanup loop

`OAS_LOCAL_QWEN_URL` is exactly the kind of pattern the loop mandate calls out:

1. **Hardcoded model name in a generic knob.** The env var hardcodes "QWEN" into an endpoint config that has nothing to do with which model the server is hosting. A deployment running Gemma or Llama on the same local endpoint would be reading an env var named after a different model family.
2. **Dual env var shape violates the "one knob per concern" principle.** `default_endpoint` was reading two env vars and picking whichever had a non-empty value. The replacement is a single knob with a clear meaning.
3. **Dead surface area in a public SDK.** OAS is a library — every env var it reads is part of its external config contract. Keeping a legacy var around just to avoid breaking a zero-user compat chain bloats that contract.

## Changes

- `lib/llm_provider/discovery.ml` — `default_endpoint` now reads only `OAS_LOCAL_LLM_URL`. The fallback chain is `OAS_LOCAL_LLM_URL → Constants.Endpoints.default_url`.
- `lib/llm_provider/discovery.mli` — doc updated.
- `lib/defaults.mli` — doc updated. The runtime binding at `defaults.ml:56` already delegates to `Discovery.default_endpoint`, so no `.ml` change is needed there.
- `test/test_defaults_unit.ml` — comment updated.

## Scope verification

`rg OAS_LOCAL_QWEN_URL` across the repo before the change returned exactly 4 matches, all of which are in this PR. No docs, no CHANGELOG, no deploy manifests, no shell scripts reference it.

## Test plan

- [x] `dune runtest test --root .` — green (all test suites pass, including `defaults_unit`)
- [x] `dune runtest lib/llm_provider --root .` — green (inline tests pass)

### DIFF
--- a/lib/defaults.mli
+++ b/lib/defaults.mli
@@ -23,8 +23,8 @@ val float_env_or : float -> string -> float
 val bool_env_or : bool -> string -> bool
 
 (** Local LLM server URL.
-    Reads [OAS_LOCAL_LLM_URL], falling back to [OAS_LOCAL_QWEN_URL],
-    then {!Llm_provider.Constants.Endpoints.default_url}. *)
+    Reads [OAS_LOCAL_LLM_URL], falling back to
+    {!Llm_provider.Constants.Endpoints.default_url}. *)
 val local_llm_url : string
 
 (** Fallback provider name.
--- a/lib/llm_provider/discovery.ml
+++ b/lib/llm_provider/discovery.ml
@@ -41,11 +41,8 @@ type endpoint_status = {
 }
 
 let default_endpoint =
-  let primary = Sys.getenv_opt "OAS_LOCAL_LLM_URL" in
-  let legacy = Sys.getenv_opt "OAS_LOCAL_QWEN_URL" in
-  match primary, legacy with
-  | Some v, _ when String.trim v <> "" -> String.trim v
-  | _, Some v when String.trim v <> "" -> String.trim v
+  match Sys.getenv_opt "OAS_LOCAL_LLM_URL" with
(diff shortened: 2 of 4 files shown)


jeong-sik:
## Cross-model review (GLM-5.1 via `sb glm-text`)

**Verdict: REQUEST_CHANGES** (accepted and addressed in commit `6922eaf`).

GLM flagged: silently dropping a documented env var in a public SDK risks silent misrouting. Even though OAS is pre-1.0 (v0.99.1) and no external references exist outside the OAS repo, the concern is valid for operator deployments.

**Resolution**: added a one-time `Printf.eprintf` migration warning at module-init time. If `OAS_LOCAL_QWEN_URL` is set, the warning fires once on process start:

```
[WARN] [Discovery] OAS_LOCAL_QWEN_URL is set (...) but has been removed
in favor of OAS_LOCAL_LLM_URL. The legacy value will be ignored.
Rename the env var to migrate.
```

The legacy URL is NOT used for routing (that would re-couple the SDK to a model name). The warning surfaces the migration mechanically. Zero cost when the env var is unset.

`dune build` + `dune runtest test` green.

PR #813, opened 2026-04-11
feat(ollama): pin keep_alive=-1 by default to prevent model eviction

jeong-sik:
## Summary

`Backend_ollama.build_request` now always sends a `keep_alive` field in the Ollama request body. Default is `"-1"` (permanent). Override via the `OAS_OLLAMA_KEEP_ALIVE` env var — any Ollama-accepted value (`"5m"`, `"30m"`, `"0"`, `"-1"`).

## Empirical rationale (2026-04-11 incident)

masc-mcp's keeper policy pins `qwen3.5:35b-a3b-nvfp4` as the resident Ollama model. An operator manually preloaded with `keep_alive=-1`. But every keeper turn that hits Ollama via OAS **reset `keep_alive` to Ollama's 5-minute default** because this backend did not set the field at all.

Within 5 minutes the model became evictable. When any concurrent process (another Claude session probe, a diagnostic script, a local test) pinged a different Ollama model, the resident `35b-a3b-nvfp4` was unloaded. Subsequent keeper requests then hit the swap window, triggering JSON truncation errors that surface downstream as:

```
Invalid request: {"error":"Value looks like object, but can't find closing '}' symbol"}
```

Counts in masc-mcp's `system_log_2026-04-11.jsonl` so far:
- **1319** occurrences of the truncation-shaped error — clustered around eviction events
- A cron-scheduled `ollama-runtime-truth.sh --repair` gate (me#984) detected drift **twice within ~1 hour** even after an operator preload

This patch is the upstream fix. Without it, even a 1-minute cron repair cadence cannot fully cover the drift window because keeper turns continuously reset the timer.

## Diff (49 +, 1 -)

- Add `keep_alive` to the non-streaming Ollama request body. Env-var-driven default is `"-1"`. Empty/whitespace env values fall back to the default.
- 2 new inline tests:
  - `build_request pins keep_alive=-1 by default`
  - `build_request honors OAS_OLLAMA_KEEP_ALIVE override`

## Test plan

- [x] `dune build --root . lib/llm_provider` — clean
- [x] `dune runtest --root .` — exit 0 (full suite)
- [x] Inline tests:
  - [x] Default: `json.keep_alive == "-1"`
  - [x] Override: `OAS_OLLAMA_KEEP_ALIVE=30m` → `json.keep_alive == "30m"`

## Scope

[...]

PR #820, opened 2026-04-11
fix(ollama): serialize keep_alive=-1 as JSON int, not string

jeong-sik:
## Summary

\`#813\` pinned \`keep_alive=-1\` by default to stop the model from being evicted — right intent. But it serialises the value as **\`String \"-1\"\`**, which Ollama's Go server parses via \`time.ParseDuration\` and rejects:

\`\`\`
Invalid request: time: missing unit in duration \"-1\"
\`\`\`

Masc-mcp's JSONL log today shows **1149 occurrences** of this exact error — every single keeper turn that cascades to \`ollama:qwen3.5:35b-a3b-nvfp4\` fails at request-parse time. The intermittent JSON-truncation bug #813 was written to prevent (\`Value looks like object, but can't find closing '}' symbol\`) has **0 occurrences** today — because the request never reaches the model at all.

Net effect of #813 on current main: turned an occasional eviction bug into a deterministic 100% failure for the Ollama cascade branch.

## Change

One branching point in \`lib/llm_provider/backend_ollama.ml\`. \`int_of_string_opt\` distinguishes:

- \`\"-1\"\`, \`\"0\"\`, \`\"300\"\` → JSON int (Ollama interprets as seconds; \`-1\` = forever)
- \`\"30m\"\`, \`\"5m0s\"\`, \`\"90s\"\` → JSON string (Ollama parses via time.ParseDuration)

Both shapes are documented Ollama API inputs, so this preserves the operator intent from #813 (\"pin permanently\") while landing on the parser branch that actually works.

\`\`\`ocaml
let keep_alive_json =
  match int_of_string_opt keep_alive_raw with
  | Some n -> \`Int n
  | None -> \`String keep_alive_raw
in
let body = (\"keep_alive\", keep_alive_json) :: body in
\`\`\`

## Tests

Updated inline \`%test\` cases in \`backend_ollama.ml\` (4 total):

- default (\`OAS_OLLAMA_KEEP_ALIVE\` unset) → JSON int \`-1\`
- default (whitespace-only env) → JSON int \`-1\`
- override \`\"30m\"\` → JSON string \`\"30m\"\`
- override \`\"300\"\` → JSON int \`300\` (new case, covers positive int path)

Verification:

- [x] \`dune build\` green
- [x] \`dune runtest lib/llm_provider/backend_ollama.ml\` green
- [ ] Post-deploy: masc-mcp log \`grep 'missing unit in duration' system_log_*.jsonl\` should drop from 1149/day to 0

## References

- #813 — initial pin keep_alive=-1 (introduced the string serialization)
- masc-mcp JSONL \`system_log_2026-04-12.jsonl\`: 1149× \"missing unit in duration '-1'\" vs 0× \"can't find closing '}' symbol\"
- Ollama API docs: \`keep_alive\` accepts bare integer (seconds) or duration string

### DIFF
--- a/lib/llm_provider/backend_ollama.ml
+++ b/lib/llm_provider/backend_ollama.ml
@@ -55,12 +55,28 @@ let build_request ?(stream=false) ~(config : Provider_config.t)
      model was evicted ~every 30 min because each keeper turn reset
      keep_alive to the 5m default; concurrent probes from other processes
      then triggered swap-induced JSON truncation errors. *)
-  let keep_alive =
+  let keep_alive_raw =
     match Sys.getenv_opt "OAS_OLLAMA_KEEP_ALIVE" with
-    | Some v when String.trim v <> "" -> v
+    | Some v when String.trim v <> "" -> String.trim v
     | _ -> "-1"
   in
-  let body = ("keep_alive", `String keep_alive) :: body in
+  (* Ollama accepts either a bare integer (seconds; -1 = forever, 0 =
+     unload immediately) or a duration string ("30m", "5m0s", ...).
+     The Go server parses string values via [time.ParseDuration], which
+     REJECTS bare integers-as-strings like "-1" with
+     [time: missing unit in duration "-1"] — the exact error we saw
+     1149 times/day in masc-mcp after oas#813 landed.
+
+     So serialise integer-parseable values as JSON ints (including the
+     negative path), and everything else as strings.  Preserves
+     compatibility for "30m"/"5m" etc. while keeping "-1" / "0" /
+     "300" working. *)
+  let keep_alive_json =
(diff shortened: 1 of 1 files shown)


jeong-sik:
## Close — superseded by #821

Same bug was fixed by a parallel session's PR #821 (merged at 5fddc7d → 782c287). The merged version of `backend_ollama.ml` already contains the `int_of_string_opt → \`Int / \`String` fix with even more complete rationale (both 2026-04-11 incidents — the 30-min eviction and the `time.ParseDuration` error).

This PR's diff is redundant and now conflicts with main. Closing in favor of #821. No loss of intent — the content of this PR's doc comment is effectively captured in #821's version.

PR #823, opened 2026-04-11
fix(builder): derive context_thresholds fallback from provider capabilities

jeong-sik:
## Summary

`Builder.with_context_thresholds` ended its resolution chain with a hardcoded `200_000` literal when none of `context_window_tokens` / `max_input_tokens` / `max_total_tokens` were set. This is the same anti-pattern that #815 addressed in `Pipeline.proactive_context_window_tokens` — a conservative default that silently under-reports the window for any model with a larger context, causing the **reducer budget** to disagree with the **compaction watermark** that #815 already derives from `Provider.capabilities_for_config`.

The disagreement matters: the pipeline decides to compact based on one view of the window, and the reducer trims to a different budget. In practice this means the reducer's target is always 200K (100K after the 0.5 compact ratio) for any agent that didn't set explicit token caps — even when the underlying provider advertises a 262K or 1M window.

## Resolution chain

Before:

1. `?context_window_tokens`
2. `b.max_input_tokens`
3. `b.max_total_tokens`
4. **`200_000` literal** ← regression target

After:

1. `?context_window_tokens`
2. `b.max_input_tokens`
3. `b.max_total_tokens`
4. **`Provider.capabilities_for_config b.provider |> max_context_tokens`** when `b.provider` is set and returns a positive value (mirrors the Pipeline#815 chain)
5. `200_000` literal as the final fallback — only reached when nothing at all is known about the model's window

## Tests

- Existing test 30 (no provider, no tokens → `200_000`) still passes because the final literal is unchanged for that scenario.
- New test 30b uses a `Local` + `qwen3-35b` config, whose `Llm_provider.Capabilities.for_model_id` returns `max_context_tokens = Some 262_144`, and verifies the reducer budget = `262_144 * 0.5 = 131_072`.

## Side observation (not fixed here)

While writing the test I hit a separate latent bug: `Provider.capabilities_for_model` for the `Anthropic` branch returns the base `anthropic_capabilities` record regardless of `model_id`, bypassing the `Llm_provider.Capabilities.for_model_id` lookup. That means `claude-opus-4-6` and `claude-sonnet-4-6` via `Provider.anthropic_{opus,sonnet} ()` currently resolve to `max_context_tokens = 200_000` in the resolver chain even though their real window is `1_000_000`.

This is why the new test uses `Local` + `qwen3-35b` instead of `Provider.anthropic_opus ()` — the Anthropic path would return the same 200K the literal fallback would, making the test unable to distinguish the branches. Worth a separate follow-up PR to make `capabilities_for_model` consult `for_model_id` for the Anthropic branch too (it already does for `Local` and `OpenAICompat`).

## Test plan

- [x] `dune exec --root . test/test_builder.exe` — 41/41 (40 existing + 1 new)
- [x] `dune runtest test --root .` — full suite green

### DIFF
--- a/lib/agent/builder.ml
+++ b/lib/agent/builder.ml
@@ -143,23 +143,14 @@ let with_context_thresholds ~compact_ratio ?context_window_tokens ?prepare_ratio
         per-agent override),
      2. builder's [max_input_tokens] (whole-run input cap),
      3. builder's [max_total_tokens] (whole-run total cap),
-     4. provider capabilities for [b.provider] when set
-        (e.g. claude-opus-4-6 → 1_000_000, glm → 200_000) — this mirrors
-        the [Pipeline.proactive_context_window_tokens] resolution chain
-        added in #815 so the reducer budget agrees with the compaction
-        watermark for the same agent.
-     5. conservative 200_000 literal as final fallback when nothing is
-        known — better to under-report than to assume a giant window
+     4. [Provider.resolve_max_context_tokens] on [b.provider] when set
+        (e.g. a [qwen3*] model_id → 262_144) — this shares the
+        "provider → capabilities → max_context_tokens" resolution with
+        [Pipeline.proactive_context_window_tokens] so the reducer budget
+        and the compaction watermark agree for the same agent.
+     5. conservative 200_000 literal as the final fallback when nothing
+        is known — better to under-report than to assume a giant window
         and skip compaction. *)
-  let from_provider () =
-    match b.provider with
-    | Some cfg ->
-      let caps = Provider.capabilities_for_config cfg in
-      (match caps.max_context_tokens with
-       | Some n when n > 0 -> n
-       | _ -> 200_000)
-    | None -> 200_000
-  in
   let effective_max = match context_window_tokens with
     | Some n when n > 0 -> n
     | _ ->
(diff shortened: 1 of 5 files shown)


jeong-sik:
## Cross-model review (GLM-5.1 via `sb glm-text`) — addressed

**Verdict**: REQUEST_CHANGES on DRY grounds (duplicated resolution chain between Pipeline #815 and this PR).

**Resolution**: commit `d4572e9` extracts `Provider.resolve_max_context_tokens : fallback:int -> config option -> int` and routes both call sites through it.

### What's shared

```ocaml
val resolve_max_context_tokens : fallback:int -> config option -> int
```

- Handles `config option` vs `config` lift
- Applies the `n > 0` guard on `caps.max_context_tokens`
- Returns `~fallback` on all failure branches

### What's deliberately *not* shared

The surrounding resolution chains in `Pipeline.proactive_context_window_tokens` and `Builder.with_context_thresholds` remain different:

- **Pipeline** (stricter): `provider → capabilities → 128K`. Explicitly rejects `max_total_tokens` / `max_input_tokens` per its docstring: "those are cumulative token budgets, not the model's per-request context-window size."
- **Builder** (more permissive): `?context_window_tokens → max_input_tokens → max_total_tokens → provider capabilities → 200K`. Accepts user-supplied token caps as hints because `with_context_thresholds` is the caller-facing setter.

So the `fallback` parameter is passed explicitly (`~fallback:128_000` from Pipeline, `~fallback:200_000` from Builder). Forcing a single constant would erase the intentional difference in semantics.

### Verification

- `dune build --root .` — clean
- `dune exec test/test_builder.exe` — 41/41
- `dune runtest test` — full suite green with the refactored helper

PR #828, opened 2026-04-11
fix(mcp): truncate_output uses CJK-aware token estimator + UTF-8 boundary

jeong-sik:
## Summary

`Mcp.truncate_output` had two correctness bugs:

1. **Over-truncation of CJK / emoji content.** The budget cap was `String.length * 4` — a byte-count approximation of the "1 token ≈ 4 characters" rule that only holds for pure ASCII. Korean/Japanese/Chinese tool output uses 3-byte characters, so against a 4-byte-per-token cap operators only got ~50% of the real character budget before the cut fired.

2. **Possible UTF-8 corruption at the truncation boundary.** `String.sub text 0 max_chars` on multi-byte content can land mid-codepoint, producing invalid UTF-8 downstream (depending on where the model reads it).

## Fix

Three changes, all in `lib/protocol/mcp.ml`:

### 1. Delegate token estimation to `Context_reducer.estimate_char_tokens`

The agent SDK already has a CJK-aware estimator in `lib/context_reducer.ml:41` with extensive test coverage (pure ASCII, pure Hangul, mixed, emoji). This PR wires `truncate_output` through it instead of rolling its own byte approximation. No new dependency — both files live in the same `agent_sdk` dune library.

### 2. Binary-search the largest prefix that fits the budget

```ocaml
let fits k =
  let safe_k = utf8_safe_boundary text k in
  let prefix = String.sub text 0 safe_k in
  Context_reducer.estimate_char_tokens prefix <= budget
in
let rec search lo hi = ...
```

For **pure-ASCII input** the estimator rounds 4 chars → 1 token, so binary search converges on exactly `budget * 4` characters — **identical to the old behavior**. The existing alcotest pins in `test_mcp_deep.ml` / `test_mcp_coverage.ml` / `test_mcp.ml` all still pass unchanged.

For **CJK input** we now keep `~budget * 1.5` characters instead of truncating at half that point. In absolute terms: for a 25K-token budget, Korean content gets ~37.5K chars reachable instead of ~12.5K.

### 3. UTF-8-safe cut point via `utf8_safe_boundary`

A small helper that walks backward from the proposed byte offset to the nearest non-continuation byte (`0x80-0xBF`). ASCII cut points are their own boundary; the worst case for well-formed input walks back 1-3 bytes.

## Test plan

Two new inline `let%test` cases pin the CJK behavior:

- `truncate_output CJK under budget is unchanged` — 9 Hangul chars, 1000-token budget, result == input
- `truncate_output CJK over budget snaps to UTF-8 boundary` — 40 Hangul chars, 2-token budget:
  - marker is present at end
  - body length is a multiple of 3 (no partial Hangul)
  - body is strictly shorter than input

Existing test suites:
- [x] `test_mcp.ml` — 23/23 alcotest cases (including `truncation 3 budget negative` edge case)
- [x] `test_mcp_deep.ml` — 24/24 (the `truncate_output 1 long "starts with 8 a's"` invariant still holds because ASCII ratio is unchanged)
- [x] `test_mcp_coverage.ml` — 22/22
- [x] `dune runtest test --root .` — full suite green

### DIFF
--- a/lib/protocol/mcp.ml
+++ b/lib/protocol/mcp.ml
@@ -20,11 +20,63 @@ type t = {
 let output_token_budget () =
   Defaults.int_env_or 25_000 "OAS_MCP_OUTPUT_MAX_TOKENS"
 
+(** Scan backward from [at] to the nearest UTF-8 codepoint boundary so a
+    byte-offset truncation never cuts the middle of a multi-byte character.
+    Returns the largest index [<= at] whose byte is NOT a UTF-8 continuation
+    byte ([0x80-0xBF]). A pure-ASCII cut point is already its own boundary;
+    the worst case for well-formed input walks back 1-3 bytes. *)
+let utf8_safe_boundary text at =
+  let len = String.length text in
+  let at = if at > len then len else at in
+  let rec scan i =
+    if i <= 0 then 0
+    else
+      let b = Char.code (String.unsafe_get text i) in
+      if b land 0xC0 <> 0x80 then i
+      else scan (i - 1)
+  in
+  scan at
+
+(** Truncate [text] when its estimated token count exceeds
+    [output_token_budget ()].
+
+    Delegates token estimation to {!Context_reducer.estimate_char_tokens}
+    so CJK / emoji content is counted on par with ASCII rather than
+    through the previous "1 token ~= 4 bytes" byte-count approximation
+    — the old formula over-truncated Korean/Japanese/Chinese tool output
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review (GLM-5.1 via sb glm-text): APPROVE. Confirms (1) binary search correctness with upper-bias mid preventing infinite loops, (2) edge cases (empty/negative/marker-only) handled cleanly, (3) O(n log n) cost is negligible vs the LLM/network I/O this gates. Full review notes in session log.

PR #830, opened 2026-04-11
fix(backend_openai): gate min_p/top_k on capabilities.supports_* (closes #827)

### LINKED ISSUE #827: backend_openai.build_request serializes min_p without checking capabilities.supports_min_p
## Symptom

Running masc-mcp JSONL shows \`oas:agent\` records with \`stop=error:Invalid request: property 'min_p' is unsupported\` for \`glm-5.1\` turns. 3 occurrences observed in a 20-record sample (15% of turns fail on this error and fall through to cascade fallback).

\`\`\`
turn=3/16 elapsed=71s duration=29s model=glm-5.1 stop=error:Invalid request: property 'min_p' is unsupported
turn=4/15 elapsed=252s duration=70s model=glm-5.1 stop=error:Invalid request: property 'min_p' is unsupported
turn=16/30 elapsed=89s duration=89s model=glm-5.1 stop=error:Invalid request: property 'min_p' is unsupported
\`\`\`

## Root cause

\`lib/llm_provider/backend_openai.ml:69-72\` serializes \`min_p\` whenever \`config.min_p\` is \`Some\`:

\`\`\`ocaml
let body = match config.min_p with
  | Some p -> (\"min_p\", \`Float p) :: body
  | None -> body
in
\`\`\`

But the \`capabilities\` type at \`lib/llm_provider/capabilities.ml:42\` already has a \`supports_min_p: bool\` field, and \`default_capabilities.supports_min_p = false\` (line 71). \`glm_capabilities\` inherits this via \`{ default_capabilities with ... }\` without overriding, so \`supports_min_p = false\` for GLM — but \`backend_openai\` doesn't consult the capability before serializing.

\`backend_glm.build_request\` delegates to \`Backend_openai.build_request\` for its base body (line 19-21), so the \`min_p\` leaks into GLM requests whenever the agent config carries an explicit \`min_p\` value (which happens via cascade inheritance or keeper defaults).

## Why \`apply_sampling_defaults\` doesn't help

\`complete.ml:39-51\` \`provider_sampling_defaults\` correctly returns \`default_min_p = None\` for \`Glm\` kind. But \`apply_sampling_defaults\` (line 58) uses the overlay pattern:

\`\`\`ocaml
min_p = (match config.min_p with Some _ -> config.min_p | None -> defaults.default_min_p);
\`\`\`

Explicit \`min_p\` set by the caller is preserved. So if the agent config carries \`min_p = Some 0.05\` (e.g., copied from a different cascade step or set as a default), it survives into the GLM request body.

## Suggested fix

\`backend_openai.build_request\`: look up the model capabilities (same lookup already done at line 79-82 for \`supports_tool_choice\`) and gate the \`min_p\` serialization:

\`\`\`ocaml
let caps = match Capabilities.for_model_id config.model_id with
  | Some c -> c
  | None -> Capabilities.openai_chat_capabilities (* or default *)
in
let body = match config.min_p with
  | Some p when caps.supports_min_p -> (\"min_p\", \`Float p) :: body
  | _ -> body
in
\`\`\`

Symmetric treatment for \`top_k\` (line 65-67) which also has a \`supports_top_k\` capability field.

## References

- masc-mcp JSONL \`system_log_2026-04-12.jsonl\` — 3× \`error:Invalid request: property 'min_p' is unsupported\` in a 20-sample window (post-#6618 Oas_log_bridge landed and is showing per-turn records from oas#816)
- OAS \`backend_openai.ml:69\`, \`capabilities.ml:42\`, \`complete.ml:58\`

jeong-sik:
## Summary

Closes #827. Running masc-mcp JSONL shows **53 occurrences** of \`oas:agent\` turn records with \`stop=error:Invalid request: property 'min_p' is unsupported\` in a ~75-minute window, with the rate accelerating as cascades retry and burn their budget on doomed glm-5.1 requests.

## Root cause

\`backend_openai.build_request\` at \`lib/llm_provider/backend_openai.ml:65-72\` serializes \`top_k\` and \`min_p\` whenever the config has \`Some _\`, without consulting the capability record. But \`Capabilities.t\` already has \`supports_top_k: bool\` and \`supports_min_p: bool\` fields, and both \`default_capabilities\` and \`glm_capabilities\` (which inherits via \`{ default_capabilities with ... }\`) have them set to \`false\`.

\`backend_glm.build_request\` delegates to \`Backend_openai.build_request\` for its base body, so every GLM request with \`config.min_p = Some _\` (inherited from cascade or agent defaults) leaked the field into the wire body and ZAI rejected it with HTTP 400.

## Fix

One \`Capabilities.for_model_id\` lookup at the top of the sampling block, gate both \`top_k\` and \`min_p\` serialization on the relevant capability. Unknown models fall back to \`default_capabilities\` — both flags \`false\`, so fields are strictly dropped unless a known model explicitly opts in.

\`\`\`ocaml
let caps =
  match Capabilities.for_model_id config.model_id with
  | Some c -> c
  | None -> Capabilities.default_capabilities
in
let body = match config.top_k with
  | Some k when caps.supports_top_k -> (\"top_k\", \`Int k) :: body
  | _ -> body
in
let body = match config.min_p with
  | Some p when caps.supports_min_p -> (\"min_p\", \`Float p) :: body
  | _ -> body
in
\`\`\`

## Tests

Three new inline \`%test\` cases:

1. **\"glm drops min_p when model does not support it\"** — sets \`~min_p:0.05\` on a \`Glm\` config, asserts the serialized JSON has no \`min_p\` key.
2. **\"glm drops top_k when model does not support it\"** — same pattern for \`~top_k:40\`.
3. **\"ollama preserves min_p (llama.cpp supports it)\"** — sanity check that providers WITH the capability still pass the field through. Guards against the fix regressing local Ollama cascades.

\`\`\`
dune build --root . — green
dune runtest --root . lib/llm_provider/backend_openai.ml — green
\`\`\`

## Observed impact

Pre-fix: **53 failures in 75 min** (0.7/min) on the live server. Each failure burns one round-trip to ZAI before cascade fallback tries the next model, so the effective keeper cascade completion rate dropped to ~1 turn per 5 minutes.

Post-fix: the request never has the unsupported field, so ZAI accepts it at the first step (glm-coding:glm-5.1 → 200). Keeper cascade completion should bounce back to ~1 turn/min baseline.

## References

- masc-mcp \`~/me/.masc/logs/system_log_2026-04-12.jsonl\`: 53× \`property 'min_p' is unsupported\` between 19:35 and 20:53
- #827 — tracking issue
- #814 (base_url log), #816 (per-turn timing), #821 (keep_alive int), masc-mcp#6618 (Oas_log_bridge) — the full observability chain that made this bug diagnosable in the first place

### DIFF
--- a/lib/llm_provider/backend_openai.ml
+++ b/lib/llm_provider/backend_openai.ml
@@ -62,13 +62,25 @@ let build_request ?(stream=false) ~(config : Provider_config.t)
     | Some p -> ("top_p", `Float p) :: body
     | None -> body
   in
+  (* Look up per-model capabilities once — the sampling params below
+     rely on [supports_top_k] / [supports_min_p] to decide whether the
+     server will accept the field.  If no capability record exists for
+     the model, default to NOT sending the non-standard sampling
+     params; conservative because misclassified providers (GLM, Gemini,
+     ...) previously silently inherited [config.min_p]/[config.top_k]
+     from the agent config and triggered hard 400s at the server. *)
+  let caps =
+    match Capabilities.for_model_id config.model_id with
+    | Some c -> c
+    | None -> Capabilities.default_capabilities
+  in
   let body = match config.top_k with
-    | Some k -> ("top_k", `Int k) :: body
-    | None -> body
+    | Some k when caps.supports_top_k -> ("top_k", `Int k) :: body
+    | _ -> body
   in
   let body = match config.min_p with
-    | Some p -> ("min_p", `Float p) :: body
-    | None -> body
+    | Some p when caps.supports_min_p -> ("min_p", `Float p) :: body
+    | _ -> body
   in
   let body = match config.enable_thinking with
     | Some enabled ->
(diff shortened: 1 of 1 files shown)


jeong-sik:
Cross-model review (GLM-5.1): APPROVE on capability gate, REQUEST_CHANGES on silent-drop observability. Addressed in commit `374be3f`: added `warn_capability_drop ~model_id ~field` with a module-level `(string * string, unit) Hashtbl.t` dedup. Fires exactly once per (model, field) pair via `Printf.eprintf`. Per-request WARN avoided — keepers at ~50 req/min on glm-5.1 now see the WARN once at first turn instead of 3000/hour. Best-effort under Eio cooperative scheduling; a race is harmless (at worst double-warn once per key). Operator-facing text points at the two actionable fixes: update `Capabilities.for_model_id` if the model actually supports the field, or remove it from the agent/cascade config.

PR #839, opened 2026-04-12
fix(ollama): disable min_p for Ollama provider (unsupported parameter)

jeong-sik:
## Summary
- Ollama rejects `min_p` with `"property 'min_p' is unsupported"`, causing keeper turns to hard-fail
- Set `supports_min_p = false` in `ollama_capabilities`
- Remove Ollama from OpenAI_compat `min_p=0.05` default — Ollama gets `None`
- OpenAI_compat (non-Ollama) still gets the 0.05 default

## Root cause
`ollama_capabilities` inherited `supports_min_p = true` from `openai_chat_extended_capabilities`. Ollama's `/api/chat` endpoint does not accept `min_p` in the `options` object.

## Test plan
- [x] Library builds clean
- [ ] Deploy and verify keeper `analyst` stops failing

### DIFF
--- a/lib/llm_provider/capabilities.ml
+++ b/lib/llm_provider/capabilities.ml
@@ -165,6 +165,7 @@ let ollama_supports_tool_choice_default =
 let ollama_capabilities = {
   openai_chat_extended_capabilities with
   supports_tool_choice = ollama_supports_tool_choice_default;
+  supports_min_p = false;
   is_ollama = true;
 }
 
--- a/lib/llm_provider/complete.ml
+++ b/lib/llm_provider/complete.ml
@@ -38,8 +38,11 @@ type sampling_defaults = {
 
 let provider_sampling_defaults (kind : Provider_config.provider_kind) : sampling_defaults =
   match kind with
-  | Provider_config.OpenAI_compat | Provider_config.Ollama ->
+  | Provider_config.OpenAI_compat ->
     { default_min_p = Some Constants.Sampling.openai_compat_min_p;
+      default_top_p = None; default_top_k = None }
+  | Provider_config.Ollama ->
+    { default_min_p = None;
       default_top_p = None; default_top_k = None }
   | Provider_config.Anthropic ->
     { default_min_p = None; default_top_p = None; default_top_k = None }


jeong-sik:
## Cross-model review (GLM-5.1 via `sb glm-text`) + source audit

**Verdict: REQUEST_CHANGES** — this PR fixes the wrong capability record.

### The error string is from GLM, not Ollama

`"property 'min_p' is unsupported"` is documented in `lib/llm_provider/backend_openai.ml:232` as the ZAI GLM response format. It is NOT Ollama's error format. GLM review confirmed independently: Ollama's open-source `api/types.go` explicitly defines `MinP float64` in its Options struct, so native Ollama cannot raise that exact string.

### Why the fix targets the wrong record

`ollama_capabilities` is only consulted when `Provider_config.kind = Ollama` — i.e. the native `/api/chat` path via `backend_ollama.build_request`. But the code there already sends `options.min_p` unconditionally:

```ocaml
(match config.min_p with
 | Some p -> options := ("min_p", `Float p) :: !options
 | None -> ());
```

So flipping `ollama_capabilities.supports_min_p = false` changes **nothing** on the native path — the serializer does not consult the capability record at all. An agent with `min_p = Some 0.05` still hits the same wire behaviour after this PR.

### The more likely failing path

The scenario that actually produces the reported error is:

1. Operator configures Ollama as **OpenAI-compat** (not native), pointing at `http://localhost:11434/v1/chat/completions`
2. `Provider_config.kind = OpenAI_compat`, `model_id = qwen3-*`
3. Cascade routes through `backend_openai.build_request`
4. `Capabilities.for_model_id "qwen3-*"` returns the qwen3 record with `supports_min_p = true`
5. `backend_openai` sends `min_p` in the body
6. Ollama's `/v1/chat/completions` endpoint (stricter than native `/api/chat`) rejects it with the error

In other words: **qwen3 via llama-server OpenAI-compat** accepts `min_p` (llama-server extended params), but **qwen3 via Ollama OpenAI-compat** does **not**. The capability is per-(server, endpoint-type) combination, not per-model.

### What this PR would actually need

One of:

1. **(a)** Route Ollama through `Provider_config.Ollama` kind (native `/api/chat`) instead of OpenAI-compat. Operators should not point at `:11434/v1/` — they should use the native `Ollama` kind. Documentation fix.
2. **(b)** In `Provider.capabilities_for_model`, detect Ollama-like `base_url` (`:11434` port, or a dedicated `is_ollama_base_url` helper) and return the qwen3 capabilities with `supports_min_p = false` overridden. Heuristic, but surgical.
3. **(c)** Add a capability gate to `backend_ollama.build_request` too, mirroring the #830/#831 pattern, so explicit user overrides also respect the declared capability. This is orthogonal to the current PR but closes the leak identified above regardless of the root cause.

### My recommendation

Close this PR. Open a new one that either:

- Adds the `backend_ollama.build_request` capability gate (option c) — **always correct**, finishes the #830/#831 series for the native path. Does not by itself fix the reported Ollama+qwen3 error, but eliminates the drift.
- Implements option (b) — a base-url-based Ollama detection in `capabilities_for_model` — which directly addresses the reported failure.

Ideally both, in sequence.

### Evidence request

Before landing anything, the author should share the actual log line with `provider_name=ollama` (or whatever provider field OAS stamps) so we can be sure the error is coming from Ollama and not a cascaded GLM fallback. After #830/#831, GLM requests should never send min_p — if the error is from GLM, it means there is a THIRD silent-drop path we haven't found yet, which is itself worth investigating.

PR #841, opened 2026-04-12
[codex] Add OAS event correlation for observability

jeong-sik:
## Summary
- add `session_id` and `worker_run_id` correlation fields to native `Event_bus` lifecycle/tool/turn events
- thread raw-trace/run metadata through agent tracing and event publication paths
- extend OAS event bus tests to cover correlated event emission

## Why
`masc-mcp` now persists and joins durable OAS events with keeper and worker telemetry. The OAS side needs stable, **MASC-agnostic** correlation metadata so raw runs, tool calls, and turn events can be stitched together without teaching OAS about MASC. The field names (`session_id`, `worker_run_id`) are generic observability terms; OAS source imports no MASC types.

## Breaking change (minor, semver-legitimate pre-1.0)

Previous framing of this PR claimed the change was "schema-additive". Cross-model review (GLM-5.1) correctly pointed out that for OCaml's type system this is **not true**. `Event_bus.event` is a variant whose constructors carry inline records:

```ocaml
type event =
  ...
  | TurnStarted of { agent_name: string; turn: int }
  | TurnCompleted of { agent_name: string; turn: int }
  ...
```

Adding fields to those inline records means every downstream pattern match must either list the new fields or add a `; _` wildcard to silence warning 9 (`unused-record-field`). External consumers building against `Event_bus.event` will need a one-line fixup.

The `examples/observable_agent.ml` diff shows exactly this:

```diff
-  | TurnStarted { agent_name; turn } ->
+  | TurnStarted { agent_name; turn; _ } ->
```

This is a **minor semver-level break** — legitimate under OAS's pre-1.0 status (currently `0.122.x`) but worth calling out so downstream rebuilds are not surprised.

### Migration

For every pattern match on `TurnStarted` / `TurnCompleted` / `ToolCalled` / `ToolCompleted`:

```ocaml
| TurnStarted { agent_name; turn } -> ...
```

becomes

```ocaml
| TurnStarted { agent_name; turn; _ } -> ...
```

Or, if you want to consume the correlation metadata:

```ocaml
| TurnStarted { agent_name; turn; session_id; worker_run_id } -> ...
```

Existing code that does not destructure these events (e.g. pattern matches that bind the whole record) is unaffected.

## Future-proofing note

A follow-up PR could wrap the two correlation fields into a single `correlation_meta` record:

```ocaml
type correlation_meta = {
  session_id: string option;
  worker_run_id: string option;
}

type event =
  | TurnStarted of { agent_name: string; turn: int; correlation: correlation_meta }
  ...
```

so future additions (`trace_id`, `span_id`, `cluster_id`) land without re-breaking the event shape. That refactor is deferred to keep this PR focused on the immediate masc-mcp correlation need.

## Impact
- OAS remains independent of MASC (verified: no `masc` import in OAS source after this PR)
- downstream consumers can join native OAS events to raw traces and worker runs
- no existing event **types** were removed; only existing event **record payloads** gained fields (minor semver break per above)

## Validation
- `dune build @all` clean
- `dune exec ./test/test_event_bus.exe` — event bus test suite green
- `dune runtest test` — full OAS test suite green


jeong-sik:
Cross-model review gate

- reviewer: direct `sb glm-text` (GLM text cascade, first model `glm-5.1`)
- reviewed_at: `2026-04-12T00:53:13Z`
- scope: PR diff vs `main`
- verdict: `No findings.`

Validation used for this branch:
- `dune build @all`
- `dune exec ./test/test_event_bus.exe`

jeong-sik:
## Cross-model review (GLM-5.1 via `sb glm-text`) — REQUEST_CHANGES

### 1. INDEPENDENCE ✓

`session_id` / `worker_run_id` are generic observability terms. No MASC types imported. The body claim holds.

### 2. BACKWARD COMPAT — critical framing issue

The PR says *"the schema change is additive"*, but for OCaml this is **not quite right**. `Event_bus.event.TurnStarted` is an inline record (`{ agent_name: string; turn: int }`) in `lib/event_bus.ml:21`. Adding fields to a record payload means every downstream pattern match on `| TurnStarted { agent_name; turn } ->` now requires `; _` to silence warning 9 (or listing the new fields). The diff at `examples/observable_agent.ml:33` literally shows this:

```diff
-  | TurnStarted { agent_name; turn } ->
+  | TurnStarted { agent_name; turn; _ } ->
```

That is the symptom of a breaking change to the event shape, not an additive one. External consumers compiling against `Event_bus.event` will need the same `_` fixup. This is a minor semver-level break.

### 3. THREADING

Surface pattern (optional params threaded through `find_and_execute_tool` / `execute_scheduled_tool`) is consistent. Can't fully verify all 13 files without tracing every call site, but nothing in the diff suggests a silent drop.

### 4. Suggested resolution

One of:

**(a) Tag as breaking.** Acknowledge the semver-minor break in the PR body + CHANGELOG. Pre-1.0 (`0.122.x`) allows it, but operators should be forewarned so downstream rebuilds pick up `; _` wildcards without surprise.

**(b) Wrap in a single `correlation` field.** Introduce a `correlation_meta = { session_id: string option; worker_run_id: string option; ... }` and add one `correlation: correlation_meta` field to each event record. Pattern matches still break once, but future correlation additions (`trace_id`, `span_id`, `masc_cluster_id`) land without breaking the shape again. **This is probably what I'd pick** — one break, stable forever.

**(c) Extensible metadata map.** `correlation: (string * string) list`. Most future-proof, least type-safe.

### Verdict

APPROVE on intent (observability + MASC-independence ✓), REQUEST_CHANGES on framing/shape. Either (a) update the body to say "minor breaking change — downstream pattern matches need `; _`" or (b) wrap the fields into a single `correlation_meta` to make this the last break to the event shape for the foreseeable future.

No semantic concern about the data itself. The timing is good — this is the kind of observability PR the /loop series has been asking for (#814 base_url + 5xx, #816 per-turn timing). Just needs honest framing or a cleaner shape.

PR #844, opened 2026-04-12
fix(backend_openai): enforce max_output_tokens capability clamp

jeong-sik:
## Summary
The capability record (`Llm_provider.Capabilities.t`) already declares `max_output_tokens` per provider/model (e.g. `openai_chat=16384`, `gpt-4o=16384`, `gpt-4.1=32000`), but neither `Backend_openai.build_request` nor `Api_openai.build_openai_body` was reading it. Both did raw passthrough of the user-supplied `max_tokens` straight into the request body, so a caller requesting more than the backend's actual cap got a 400 at the final-response step — **after** any mutating tools in the turn had already committed.

The abstraction was declared but not enforced — a fake implementation in the strict sense.

## Evidence — live keeper stall (2026-04-12)

MASC keeper `analyst` running the `keeper_unified` cascade (glm-coding → glm → **groq** → ollama) with `max_tokens: 65536`. Groq's `qwen/qwen3-32b` has a 40960 max-output cap that is surfaced only via the 400 response, not via any capability record we consult.

```
2026-04-12T00:37:49Z | keepalive turn scheduled for analyst: channel=scheduled_autonomous
2026-04-12T00:38:06Z | turn completed
2026-04-12T00:38:06Z | oas_worker: agent errored (no proof):
    Invalid request: `max_tokens` must be less than or equal to `40960`,
    the maximum value for `max_tokens` is less than the `context_window`
2026-04-12T00:38:06Z | analyst: unified turn FAILED cascade=keeper_unified
    max_context=262144 latency=16297ms
2026-04-12T00:38:06Z | registry: phase transition name=analyst
    old=running new=failing
    event=manual_reconcile_required(ambiguous_partial_commit(
      post_commit_failure:
      Mutating tools [keeper_board_comment] committed before the turn failed;
      retry stayed disabled and manual reconcile is required))
```

The keeper had already committed `keeper_board_comment` to state before the LLM final-response hit the 400. MASC's safety mechanism cannot safely retry (would double-commit) and opens a `manual_reconcile` sidecar. Across the full 6-keeper cohort this reliably cascades into a room-wide stall (`analyst`, `cheolsu`, `janitor`, `masc-improver`, `sangsu`, `uranium666` all hit the same pattern with different `committed_tools` — `masc_add_task`, `keeper_board_vote`, `masc_cleanup_zombies`, etc.).

## Fix

Both request builders now apply a ceiling clamp against `capabilities.max_output_tokens`:

- `lib/llm_provider/backend_openai.ml::build_request` — moves the existing `caps = ...` lookup earlier in the function and clamps `max_tokens` before adding it to the body. Removes the duplicate lookup later.
- `lib/api_openai.ml::build_openai_body` — uses the already-computed `capabilities` variable (line 68) to clamp before building the body.

Both paths call a new helper `Llm_provider.Backend_openai.warn_capability_clamp` — distinct from `warn_capability_drop` because the semantic is **value-lowering**, not **field-dropping**. Shares the same dedup table but uses a `:clamp` sentinel so a clamp warning does not silence a subsequent drop warning on the same `(model_id, field)` pair.

`[None]` cap semantics: "unknown backend" → pass through, rather than imposing an arbitrary ceiling. `[Some cap]`: lower oversized requests to cap, emit one-shot stderr WARN.

## No hardcoded numbers

The clamp reads the number from `capabilities.max_output_tokens` only. There is **no new hardcoded integer** anywhere in `backend_openai.ml` or `api_openai.ml`. All caps live in `capabilities.ml` where they already did — this PR only enforces the existing abstraction.

If a specific provider's cap is stale (e.g. Groq's qwen/qwen3-32b actually caps at 40960 but currently inherits `openai_chat_capabilities.max_output_tokens = Some 16_384` via `default_openai_compat_capabilities`), that is a **separate capability-accuracy fix** and out of scope for this PR. This PR ensures that whatever value is declared in `capabilities.ml` is actually enforced at request time. A follow-up can add `groq_capabilities` or a `starts_with \"qwen/qwen3-32b\"` override once the Groq team publishes authoritative per-model caps.

## Tests
Three new test cases in `test/test_api.ml::build_body_assoc`:

1. **clamp applied**: `gpt-4o` (cap=16384) + `max_tokens=32768` → body shows `16384`
2. **passthrough within cap**: `gpt-4o` + `max_tokens=4096` → body shows `4096` (unchanged)
3. **default-compat clamp for unknown model**: `totally-unknown-model-2099` + `max_tokens=65536` → body shows `16384` (falls through to `default_openai_compat_capabilities` = `openai_chat_capabilities.max_output_tokens = Some 16384` — the abstraction's deliberate safety default)

Full suite: 64 tests in `test_api` (was 61), all 30+ OAS test suites pass with 0 regressions.

## Observability change

Clamp now emits a one-shot stderr WARN:
```
[WARN] [backend_openai] clamping max_tokens for model <model>: requested <N>
  exceeds capability cap <M>. Update Capabilities.for_model_id if the advertised
  cap is stale, otherwise lower the value in your agent/cascade config to avoid
  the silent truncation.
```
Operators will see this ONCE per `(model_id, \"max_tokens\")` pair on the first oversized request — it is a debugging hint, not a per-request warning.

## Test plan
- [x] `dune build --root .` clean
- [x] `dune runtest --root .` 30+ suites green
- [x] `test_api` 64 tests (+3 new)
- [ ] Cross-model review (GLM-5.1)
- [ ] Merge → OAS release
- [ ] masc-mcp OAS pin bump → worktree rebuild → live restart → verify MASC keeper sidecars no longer open with `max_tokens` 400 root cause

## Related
- `#839` (fix(ollama): disable min_p for Ollama provider) — sibling capability-accuracy fix (different field)
- MASC cycle-25 keeper stall cohort, full 6/6 reproduce

### DIFF
--- a/lib/api_openai.ml
+++ b/lib/api_openai.ml
@@ -70,11 +70,28 @@ let build_openai_body ?provider_config ~config ~messages ?tools ?slot_id () =
     system_message_json config
     @ List.concat_map openai_messages_of_message messages
   in
+  (* Clamp [max_tokens] to the provider's advertised [max_output_tokens]
+     ceiling, if declared. Rationale: raw passthrough of a user-supplied
+     [max_tokens] that exceeds the backend's cap produces a 400 error
+     after the turn has already committed mutating tools (observed on
+     MASC keepers against Groq's qwen/qwen3-32b at 40960), leaving the
+     caller in an [ambiguous_post_commit_failure] state. [None] means
+     "unknown cap" → pass through. Parallel implementation to the one
+     in [Llm_provider.Backend_openai.build_request]. *)
+  let effective_max_tokens =
+    match capabilities.max_output_tokens with
+    | Some cap when config.config.max_tokens > cap ->
+        Llm_provider.Backend_openai.warn_capability_clamp
+          ~model_id:model_str ~field:"max_tokens"
+          ~requested:config.config.max_tokens ~cap;
+        cap
+    | _ -> config.config.max_tokens
+  in
   let body_assoc =
     [
       ("model", `String model_str);
       ("messages", `List provider_messages);
-      ("max_tokens", `Int config.config.max_tokens);
+      ("max_tokens", `Int effective_max_tokens);
     ]
   in
   let body_assoc =
--- a/lib/llm_provider/backend_openai.ml
+++ b/lib/llm_provider/backend_openai.ml
@@ -47,6 +47,24 @@ let warn_capability_drop ~model_id ~field =
       field model_id field
   end
 
+(* Ceiling-based clamp (used for [max_tokens] against
+   [capabilities.max_output_tokens]). Semantics differ from
+   [warn_capability_drop] — the field is NOT dropped, the value is
+   lowered to the advertised upper bound. Shares the dedup table but
+   uses a distinct sentinel so a clamp does not silence a subsequent
+   drop on the same (model, field) pair. *)
+let warn_capability_clamp ~model_id ~field ~requested ~cap =
+  let key = (model_id, field ^ ":clamp") in
+  if not (Hashtbl.mem capability_drop_warned key) then begin
+    Hashtbl.replace capability_drop_warned key ();
+    Printf.eprintf
+      "[WARN] [backend_openai] clamping %s for model %s: requested %d \
+       exceeds capability cap %d. Update Capabilities.for_model_id if \
+       the advertised cap is stale, otherwise lower the value in your \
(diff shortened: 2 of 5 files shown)


jeong-sik:
Closing — superseded by #846 which unifies the two redundant `Capabilities.for_model_id` lookups into one `caps` binding (the max_tokens clamp and the top_k/min_p gates now share a single capability lookup). #846 is based on current main with no conflicts.

PR #845, opened 2026-04-12
feat(event_bus): wrap events in envelope for causal correlation

jeong-sik:
## Summary

Closes the Message Bus correlation gap flagged by Anthropic's [Multi-Agent Coordination Patterns](https://claude.com/blog/multi-agent-coordination-patterns) guide.

Before this change, `Agent_sdk.Event_bus.event` variants carried no causal metadata. A `ToolCalled` observed on the SSE stream could not be joined back to its originating turn or task — the `(agent_name, turn: int)` pair is not unique for concurrent fibers, and most variants (`ToolCalled`, `ToolCompleted`, `TurnStarted`, `TurnCompleted`, `ContextCompacted`) carried no `task_id` at all. Consumers (MASC `lib/oas_sse_bridge.ml`, external dashboards) could not synthesize the missing IDs client-side because `publish` ordering is fiber-interleaved and non-deterministic.

### Phase 1 (this PR)

```ocaml
type envelope = {
  correlation_id : string;  (* stable across a causal chain *)
  run_id : string;          (* unique per physical Agent.run *)
  ts : float;               (* emission time, not publish time *)
}

type event = { meta : envelope; payload : payload }

val fresh_id : unit -> string
val mk_envelope : ?correlation_id:string -> ?run_id:string -> unit -> envelope
val mk_event : ?correlation_id:string -> ?run_id:string -> payload -> event
val filter_correlation : string -> filter
val filter_run : string -> filter
```

- **Orchestrator** uses `task.id` as `correlation_id` for `AgentStarted` / `AgentCompleted`.
- **Pipeline** uses `agent.lifecycle.current_run_id` (already driven by `Raw_trace.active_run_id`) for `TurnStarted` / `TurnCompleted` / `ContextCompacted` so all events within one `Agent.run` share a chain.
- **Agent_tools** threads `?current_run_id` through `execute_tools` / `execute_scheduled_tool` / `find_and_execute_tool` so `ToolCalled` / `ToolCompleted` within one turn share a chain.
- **Event_forward** JSON output gains `correlation_id` / `run_id` columns so file sinks carry the causal chain for offline joining.

### Phase 2 (follow-up, not in this PR)

- Context.t plumbing so **orchestrator → pipeline** correlation flows through (currently each layer generates its own chain).
- TLA+ causality spec: every `ToolCalled` has a preceding `TurnStarted` with the same `run_id`.
- Consumer side (MASC `oas_sse_bridge.ml`) emits envelope fields in the SSE JSON.

## Tradeoffs

- **Breaking API change**. `type event` is now a record; pattern matches must use `event.payload`. Construction must go through `Event_bus.mk_event` or explicit record literal. All downstream call sites in `lib/`, `test/`, and `examples/observable_agent.ml` have been updated in this PR.
- **`fresh_id` is not a UUID**. Format `<pid-hex>-<micros-hex>-<counter-hex>` uses an atomic counter for process-local uniqueness and a PID prefix for cross-process uniqueness. No new dependency; collision bound is acceptable for offline correlation. The 1000-id uniqueness test confirms this in practice.
- **Phase 1 does not yet thread correlation across layers**. Orchestrator and pipeline each generate their own chain. The handoff requires Context.t surgery that's worth a separate review cycle.
- **`fields.current_run_id` fallback**. When `Agent.run` is invoked without Raw_trace (no `current_run_id` set), pipeline falls back to a fresh per-event id, which breaks joining across events within that run. This matches the existing Raw_trace gating — not a regression.

## Test plan

- [x] `dune exec test/test_event_bus.exe` — 24 tests (19 preexisting + 5 new envelope tests)
  - `fresh_id_unique`: 1000 ids, all distinct
  - `mk_envelope_defaults`: explicit and auto-generated fields
  - `filter_correlation`: 2 interleaved chains, only 3 events match chain-a
  - `filter_run`: 2 runs in same chain, only run-2 matches
  - `envelope_ts_monotonic`: construction-time ts (not publish-time) preserves order
- [x] `dune exec test/test_event_forward.exe` — 20 tests
- [x] `dune exec test/test_orchestrator.exe` — 73 tests
- [x] `dune exec test/test_elicitation.exe` — 12 tests
- [x] `dune exec test/test_deep_coverage.exe` — 83 tests
- [x] Full `dune build` green

## Breaking change migration

| Before | After |
|--------|-------|
| `match event with \| TurnStarted r -> ...` | `match event.payload with \| TurnStarted r -> ...` |
| `Event_bus.publish bus (TurnStarted {...})` | `Event_bus.publish bus (Event_bus.mk_event (TurnStarted {...}))` |
| `Event_forward.agent_name_of_event e` | `Event_forward.agent_name_of_payload e.payload` |
| `type event = AgentStarted \| ...` (as `Event_bus.event`) | `type payload = AgentStarted \| ...` (+ `type event = { meta; payload }`) |


jeong-sik:
#841 (inline correlation fields) was merged while this PR was in flight. Since #845 proposes a different structural approach (envelope wrapper) for the same Event_bus.event types, the two changes conflict. Version Consistency also fails (likely stale version after the v0.122.1 bump). Options: (a) rebase on main and adapt the envelope approach to wrap around #841's merged inline fields, or (b) start fresh from current main. Given the 22-file scope, (b) is probably cleaner. The correlation_meta wrapper idea from my review on #841 is still a valid follow-up — it just needs to build on top of #841's merged state rather than replacing it.

PR #850, opened 2026-04-12
feat(provider): make max_tokens optional — None defers to model capability

jeong-sik:
## Summary

`Provider_config.t.max_tokens`, `Types.agent_config.max_tokens`, and `Builder.t.max_tokens` are now `int option` instead of `int`.

- `None` = "defer to model capability" — resolved at request time via `Capabilities.for_model_id`
- `Some n` = "caller override, clamped to model cap when known"

## Why

masc-mcp hardcoded `keeper_unified_max_tokens = 65536` (mislabeled "OpenHands default" — OpenHands actually defaults to None/model-specific per issue #13565). GLM coding caps at 40960, Groq at 40960. The 65536 caused HTTP 400 → `ambiguous_partial_commit` → `manual_reconcile` deadlock on every keeper turn.

Emergency patches in the last 24h:
- #6681 removed providers from cascade
- #6687 lowered to 32768
- #6693 lowered to 16384

Each was an arbitrary number with no empirical basis. The correct fix: stop hardcoding, let the capability system resolve.

## Resolution logic (backend_openai.ml)

```ocaml
match config.max_tokens, caps.max_output_tokens with
| None, Some cap -> cap      (* defer → model ceiling *)
| None, None     -> 4096     (* unknown model fallback *)
| Some n, Some cap when n > cap -> cap  (* clamp + warn *)
| Some n, _      -> n        (* explicit, pass through *)
```

Other backends (Anthropic, Gemini, Ollama) use `Option.value ~default:4096` since their APIs require the field.

## API change

```ocaml
(* Before *)
Builder.with_max_tokens : int -> t -> t  (* mandatory *)

(* After *)
Builder.with_max_tokens : int -> t -> t  (* wraps in Some internally *)
(* Not calling with_max_tokens = None = dynamic resolution *)
```

Callers that want "use the model's own ceiling" simply don't call `with_max_tokens`. The default is `None`.

## Scope

36 files (19 lib + 2 examples + 15 tests). All mechanical `int → int option` propagation. Core semantic change is in `backend_openai.ml` (4-branch resolution) and `Provider_config.make` (default None).

## Test plan

- [x] `dune build --root .` → 0 errors
- [x] `dune runtest --root .` → all tests pass
- [ ] CI green

### DIFF
--- a/examples/cascade_failover.ml
+++ b/examples/cascade_failover.ml
@@ -51,7 +51,7 @@ let () =
      Printf.printf "   temperature: %s\n"
        (match cfg.temperature with
         | Some t -> Printf.sprintf "%.1f" t | None -> "default");
-     Printf.printf "   max_tokens: %d\n" cfg.max_tokens
+     Printf.printf "   max_tokens: %s\n" (match cfg.max_tokens with Some n -> string_of_int n | None -> "auto")
    | None ->
      Printf.printf "   (not available)\n");
 
--- a/examples/streaming.ml
+++ b/examples/streaming.ml
@@ -41,7 +41,7 @@ let () =
       default_config with
       model = provider.model_id;
       system_prompt = Some "You are a helpful assistant.";
-      max_tokens = 1024;
+      max_tokens = Some 1024;
     }
   in
   let messages =
--- a/lib/agent/builder.ml
+++ b/lib/agent/builder.ml
@@ -8,7 +8,7 @@ type t = {
   model: model;
   name: string;
   system_prompt: string option;
-  max_tokens: int;
+  max_tokens: int option;
   max_turns: int;
   temperature: float option;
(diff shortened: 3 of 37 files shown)


jeong-sik:
Cross-model review update for head `6049ed689cf647751618cce2d7796da17fd89649`

- Model: direct ZAI `sb glm-text --no-thinking`
- Timestamp: 2026-04-13 11:06:28 KST
- Result: `No findings.`
- Note: I reran the review against the final diff after rebase to avoid false positives from intermediate commits that will be squashed away.

Local verification:
- `opam exec -- dune runtest --root . lib/llm_provider`
- `opam exec -- dune exec --root . ./test/test_types.exe`
- `opam exec -- dune exec --root . ./test/test_builder.exe`

PR #852, opened 2026-04-12
fix(ollama): send tool_call arguments as JSON object, not string

jeong-sik:
## Summary
- Ollama `/api/chat` expects `tool_calls[].function.arguments` as a JSON **object**, but OAS was sending it as a JSON **string** (via `Yojson.Safe.to_string`)
- This caused Ollama's Go template parser to fail with `"Value looks like object, but can't find closing '}' symbol"` on every multi-turn keeper interaction after the first tool call
- Root cause of **404+ VLLO errors/day** and **6 keeper stall** in masc-mcp

## Changes
- Add optional `~arguments_as_object` parameter to `tool_calls_to_openai_json` and `openai_messages_of_message` (default `false` for OpenAI/GLM backward compat)
- `backend_ollama.ml` passes `~arguments_as_object:true`
- Re-exports in `backend_openai.ml` eta-expanded to hide the optional parameter from public API
- 2 new inline tests verifying both formats

## Evidence
- Direct Ollama test with `arguments` as string: rejected with "can't find closing '}'"
- Direct Ollama test with `arguments` as object: accepted
- Ollama issues: #14570, #14493 (Qwen 3.5 tool calling)
- All existing OAS tests pass

## Test plan
- [x] `dune build` passes
- [x] `dune runtest` passes (all existing + 2 new tests)
- [ ] Deploy to masc-mcp and verify VLLO count drops to 0

### DIFF
--- a/lib/llm_provider/backend_glm.ml
+++ b/lib/llm_provider/backend_glm.ml
@@ -37,6 +37,13 @@ let build_request ?(stream=false) ~(config : Provider_config.t)
             ("thinking", thinking) :: fields
         | None -> fields
       in
+      (* Remove chat_template_kwargs leaked from Backend_openai.
+         GLM does not recognize this llama.cpp/Ollama-specific field and
+         returns "Invalid API parameter" when it is present.  GLM thinking
+         is handled above via the native [thinking] parameter. *)
+      let fields =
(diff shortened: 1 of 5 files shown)


jeong-sik:
Superseded by #848 which already includes the Ollama tool_calls serializer fix (tool_calls_to_ollama_json + ollama_messages_of_message).

PR #929, opened 2026-04-15
docs: sync swarm execution example with latest API

jeong-sik:
Sync README.md swarm execution example with Swarm_types and Runner.run signature. Remove outdated version string.


jeong-sik:
Superseded by #934, which removes the deprecated swarm documentation instead of refreshing the old example.

PR #1047, opened 2026-04-19
[codex] Fix Gemini CLI empty MCP whitelist

jeong-sik:
## Summary

- Replace the Gemini CLI MCP-off empty whitelist argument with a non-matching sentinel name.
- Keep explicit `OAS_GEMINI_ALLOWED_MCP` allow-lists unchanged.
- Update inline tests and changelog coverage for Gemini CLI 0.38 policy validation.

## Root Cause

Gemini CLI 0.38 rejects `--allowed-mcp-server-names ""` with `mcpName is required if specified (cannot be empty)`. OAS 0.162.0 used that empty value for the default MCP-off path and for `OAS_GEMINI_NO_MCP=1`, so MASC keeper cascades could fail before the model call started.

## Validation

- `dune runtest --root . lib/llm_provider`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ original tag dates. `0.100.4` was never tagged or released.
 
 ### Fixed
 
+- **Gemini CLI MCP-off no longer emits an empty MCP server name.** Gemini CLI
+  0.38 rejects `--allowed-mcp-server-names ""` with
+  `mcpName is required if specified (cannot be empty)`. The `gemini_cli`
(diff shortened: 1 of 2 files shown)


jeong-sik:
Superseded by #1048, which landed the same Gemini CLI MCP sentinel fix on `main` as `54f4aeabd0ff88b9666d1c3ecc35aee26dbc894f`.

Closing this PR to avoid carrying a duplicate/conflicting implementation branch.

PR #1122, opened 2026-04-21
feat(provider_kind): typed provider_kind across inference_telemetry

jeong-sik:
## Summary

Lifts `inference_telemetry.provider_kind` from a free-form `string option` to a typed sum, while keeping the on-disk / over-the-wire JSON payload byte-for-byte identical. Three commits tell the story:

1. `feat(provider_config): add provider_kind_of_string canonical parser` — inverse of the existing `string_of_provider_kind`; accepts canonical lowercase + documented aliases (`claude`, `openai`, `llama`); case-insensitive, whitespace-trimmed; unknown input returns `None`.
2. `feat(provider_config): add show + yojson serializers for provider_kind` — hand-written so the wire form stays `"anthropic"` rather than the capitalised `"Anthropic"` that `[@@deriving yojson]` would emit by default.
3. `feat(types): lift inference_telemetry.provider_kind to the typed sum` — setters write `Some config.kind`; readers pattern-match on the constructor; a new module `Provider_kind` holds the type so it can be shared by `Types` without creating a `Provider_config <-> Types` dependency cycle. `Provider_config` rebinds the type + helpers so every existing `Provider_config.Anthropic` caller continues to compile unchanged.

## Why

- The codebase already has an exhaustive `string_of_provider_kind`, but there was no matching parser. That asymmetry has produced three drift bugs today:
  - `provider_bridge.ml:46` accepts `"llama"` and `"ollama"` for the same kind; `string_of_provider_kind Ollama` only emits one of them.
  - `provider_bridge.ml:93–102` serialises `Anthropic → "claude"`, while `string_of_provider_kind Anthropic = "anthropic"` — two different conventions live in parallel.
  - `agent_config.ml:224–227` handles `"anthropic"` and `"openai"` but not `"openai_compat"`, which is the canonical string the SDK itself emits.
- `inference_telemetry.provider_kind` setters were already stringifying a typed value only to have reader sites reparse it ad-hoc — `t.provider_kind = Some "ollama"` pattern. This PR removes that dance.

## Scope

- Additive for the first two commits; structural for the third (type change + module split).
- Wire format is preserved: `inference_telemetry` JSON still contains `"provider_kind":"anthropic"` etc. A new `telemetry_wire_format` test group in `test_provider_config.ml` asserts this for every variant and closes with a full JSON round-trip.
- No caller-sweep of `provider_bridge.ml` / `agent_config.ml` / `provider.ml` ad-hoc ladders — that is a separate review concern and lands as a follow-up PR.

## Verification

- `dune build --root .` green.
- `dune exec --root . test/test_provider_config.exe` — **37 tests** (20 pre-existing + 5 parser + 7 serializer + 3 telemetry wire-format), all green.
- `dune exec --root . test/test_provider_complete.exe` — **27 tests**, all green. The migrated `provider_kind placeholder` Alcotest uses a local testable built from `pp_provider_kind`.
- `dune build --root . @test/runtest` — full oas test suite green, including the runtime smoke and multi-vendor live tests.

## Files touched

| File | Change |
|------|--------|
| `lib/llm_provider/provider_kind.ml{,i}` | **New module** owning the sum type + all helpers. |
| `lib/llm_provider/provider_config.ml{,i}` | Rebinds type + helpers from `Provider_kind`; implementation is now a one-liner delegation. |
| `lib/llm_provider/types.ml{,i}` | `inference_telemetry.provider_kind : string option → Provider_kind.t option`. |
| `lib/llm_provider/complete.ml` | Setter drops `string_of_provider_kind`; two readers pattern-match on the constructor. |
| `lib/api.ml` | One literal `Some "anthropic"` in a `let%test` becomes `Some Llm_provider.Provider_config.Anthropic`. |
| `test/test_provider_complete.ml` | Local `provider_kind_t` testable for the one affected Alcotest check. |
| `test/test_provider_config.ml` | New `telemetry_wire_format` group (+3 cases) plus the already-landed parser / serializer groups. |

## Test plan

- [x] `dune build --root .`
- [x] `dune exec --root . test/test_provider_config.exe` (37/37)
- [x] `dune exec --root . test/test_provider_complete.exe` (27/27)
- [x] `dune build --root . @test/runtest`
- [ ] CI runs
- [ ] Follow-up PR: migrate scattered string ladders in `provider_bridge.ml`, `agent_config.ml`, `provider.ml`, `agent_lifecycle.ml`, `protocol/agent_card.ml` to consume `Provider_kind.of_string` so the compiler catches every new-provider addition at every site.


jeong-sik:
Added a follow-up commit on the same branch: hand-written `show` + `yojson` serializers for `provider_kind` so records that embed it (e.g. `Types.inference_telemetry`) can switch from `string option` to `provider_kind option` without changing the wire format. 7 new tests added; the provider_config suite now has 34/34 passing. The migration itself will land in a follow-up PR.

PR #1165, opened 2026-04-24
fix(fs): unique tmp per writer to close atomic-write rename race

jeong-sik:
## Root cause (one-liner)

Atomic file writes used a target-derived shared tmp name (`<path>.tmp`), so two fibers writing to the same logical file could race: writer A's `rename` consumes the shared tmp before writer B's `rename` runs, producing `Eio.Io Fs Not_found (renameat …)`.

## Fix

- **`lib/fs_atomic_eio.ml` (new)**: `save_atomic` helper for `Eio.Path`. Writer-unique tmp suffix (pid + wall-clock ns + `Atomic` counter), best-effort `fsync` of tmp and parent directory, `Cancelled` re-raised after tmp cleanup.
- **`lib/fs_result.ml`**: same discipline applied to the blocking `write_file` path via `Filename.temp_file`, `Out_channel.flush`, `Unix.fsync`.
- **5 call sites migrated**: `checkpoint_store.save`, `a2a_task_store.store_task`, `memory_file_backend.persist`, `durable_event.save_to_file`, and the now-unified `Fs_result.write_file`. The in-memory cache in `a2a_task_store` is only updated on `Ok`, so cache/disk can no longer drift if the atomic write fails after flush.
- **List-filters widened**: `.json.tmp` (9-char) suffix → `.tmp` (4-char) in `checkpoint_store` / `a2a_task_store` / `memory_file_backend`, because unique suffixes no longer end in exactly `.json.tmp`.

## Proof

`test/test_atomic_write_race.ml` (4 cases, Alcotest + Eio):

1. `save_atomic` round-trip byte equality.
2. repeated saves pick last writer's content.
3. **8 fibers save to the same name concurrently** — every call returns `Ok`, target contains one writer's bytes, no `*.tmp` leftover. This case would fail with `renameat ENOENT` against the old shared-tmp code.
4. 8 fibers save to distinct names — each target has its own content.

Full `dune runtest` green (exit 0) including `test_checkpoint_store`, `test_a2a_task_store`, `test_memory_file_backend` unchanged.

## Related (downstream)

The shared-tmp rename race was observed in MASC keeper persistence paths that sit on top of OAS helpers:

- Related: jeong-sik/masc-mcp#9749 (fsync gap between rename and write-back)
- Related: jeong-sik/masc-mcp#9769 (checkpoint write_meta CAS conflict symptoms)
- Related: jeong-sik/masc-mcp#9780 (renameat ENOENT under concurrent save)

A separate downstream PR will bump the OAS pin in masc-mcp after this merges.

## Version

0.170.6 → 0.170.7 via `scripts/sync-version-truth.sh --apply` (3-axis SSOT).


jeong-sik:
Build & Test 재실행도 동일 실패 — `io_uring ENOMEM` 인프라 이슈로 판정. 본 PR diff는 원인 아님 (새 test는 BISECT_ENABLE 가드로 coverage 중 skip). 별도 이슈 등록 필요 — 이를 해결하기 전엔 merge 차단됨.

자세한 내용은 CI 인프라 이슈 트래킹 중. 본 PR은 인프라 fix 선행 후 재가동 권장.

PR #1309, opened 2026-05-01
ci: trigger draft skip on draft conversion

jeong-sik:
## Summary
- include `converted_to_draft` in the CI pull_request trigger
- let the existing concurrency key cancel an in-flight ready PR CI run when the PR is moved back to draft

## Validation
- git diff --check HEAD~1..HEAD
- ruby YAML parse for .github/workflows/ci.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-    types: [opened, synchronize, reopened, ready_for_review]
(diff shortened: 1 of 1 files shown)


jeong-sik:
Follow-up for the review feedback resolved on #1308 after it merged: this carries the same converted_to_draft trigger fix on top of current main. Validation re-run: git diff --check HEAD~1..HEAD; ruby YAML parse for .github/workflows/ci.yml.

PR #1352, opened 2026-05-04
refactor(capabilities): data-driven for_model_id lookup

jeong-sik:
## Summary
- Replace 360-line `if-else` prefix chain in `for_model_id` with ordered `model_table` lookup
- Extract shared capability records (`deepseek_v4_capabilities`, `glm_fast_capabilities`, `glm_text_capabilities`) to reduce duplication
- Add `for_model_dynamic` for models needing runtime inspection (nemotron VL sub-prefix, gemma-4 27B+ audio sub-prefix)
- Net: -40 LOC (283 added, 323 removed)

## Why
AI-generated code created a sequential if-else chain by appending new model entries without refactoring. Each new model added ~10 lines to an already 360-line function. Data-driven table is easier to maintain, audit, and extend.

## Anti-pattern: String Matching via if-else chain
Prefix matching via cascading `if ... then ... else if ...` is O(n) in model count with no structure. Ordered list with `List.exists` is semantically identical but data-separated.

## Semantics preserved
- Match order: table entries are ordered identically to original if-else
- Fallback: unknown models still return `None` (caller applies `default_capabilities`)
- Dynamic inspection: nemotron/gemma-4 special cases extracted to `for_model_dynamic`

## Test plan
- [x] `dune runtest` — 37/41 pass (4 pre-existing emits_usage_tokens failures, same as before refactoring)
- [x] No new test failures introduced

### DIFF
--- a/lib/llm_provider/capabilities.ml
+++ b/lib/llm_provider/capabilities.ml
@@ -368,206 +368,285 @@ let codex_cli_capabilities =
 
 (* ── Model-specific overrides (lookup table) ─────────── *)
 
-(** Lookup capabilities by model_id prefix.
-    Returns None if no specific override is known. *)
-let for_model_id model_id =
-  (* Normalize: lowercase, strip quantization suffixes *)
-  let m = String.lowercase_ascii model_id in
-  let starts_with prefix =
-    String.length m >= String.length prefix
-    && String.sub m 0 (String.length prefix) = prefix
-  in
-  if starts_with "claude-opus-4"
-  then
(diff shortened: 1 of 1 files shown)


jeong-sik:
Superseded by #1337 which added Nemotron/Gemma 4/GLM-5-Code entries directly to the for_model_id chain. The model_table refactoring is valuable but should be re-based from current main (0.188.0) in a future PR to avoid merge conflict risk with the 35 new entries.

PR #1356, opened 2026-05-04
test: add regression tests for capability-gated serializer anti-patterns (P5)

jeong-sik:
## Summary
- Add 4 inline regression tests in `backend_openai.ml` for capability-gated serializer paths (max_tokens clamping, seed omission, Chat_template_kwargs, pass-through)
- Add 7 inline regression tests in `backend_ollama.ml` for think parameter, num_predict mapping, top_k gate, stop_reason mapping, error parsing, thinking block extraction

## Anti-pattern coverage
These tests guard against regressions in 8 of the 32 anti-patterns identified in the Kimi/GLM compatibility analysis:
- **S01/S02**: `warn_capability_drop` gate for top_k/min_p
- **S03**: `thinking_control_format` capability-based branching
- **S07**: max_tokens capability ceiling clamping
- **H06**: keep_alive integer/string wire format (already covered by 7 existing tests)
- **H07**: think default=false, explicit true/false from enable_thinking
- **M02**: deepseek thinking via capability field (not string matching)

## Test plan
- [x] `dune build @install` passes in worktree
- [x] `dune runtest lib/llm_provider/backend_openai.ml` — all tests green
- [x] `dune runtest lib/llm_provider/backend_ollama.ml` — all tests green (21 total)

### DIFF
--- a/lib/llm_provider/backend_ollama.ml
+++ b/lib/llm_provider/backend_ollama.ml
@@ -688,12 +688,15 @@ let%test "build_request maps max_tokens to num_predict in options" =
     json |> member "options" |> member "num_predict" |> to_int = 2048)
 ;;
 
-let%test "build_request omits top_k from options when capability says no" =
-  (* ollama_capabilities has supports_top_k=true by default.
-     To test the capability drop path, we need a model that resolves
-     to capabilities with supports_top_k=false. Since we can't easily
-     inject custom capabilities, test the default path (top_k included)
-     and verify the code path exists. *)
+let%test
+    "build_request includes top_k in options when supports_top_k=true (default ollama \
+     caps)"
+  =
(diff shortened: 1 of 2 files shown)


jeong-sik:
Addressed Copilot review threads in 0d0cf335:

1. **`backend_ollama.ml` top_k test naming** — Renamed the test to reflect what it actually exercises (`includes top_k in options when supports_top_k=true (default ollama caps)`) and added a comment pointing to `Backend_openai`'s gate as the canonical drop-path coverage.

2. **`String.sub msg 0 5` length guard** — Replaced with `String.starts_with ~prefix:"model"`, which removes the partial-substring assertion entirely.

`ocamlformat --auto-promote` also reflowed two adjacent lines.

Local `dune build lib` is clean. Leaving Draft until human approves Ready.

PR #1357, opened 2026-05-05
fix(streaming): surface SSE parse failures instead of silent discard

jeong-sik:
## Summary

\`parse_sse_event\` previously returned \`None\` for two distinct failure modes — unknown event type and JSON parse error. The downstream accumulator treated \`None\` the same as \"no event in this chunk\" (a legitimate heartbeat-shaped payload), so the failure never surfaced. The stream then accumulated to \`MessageStop\` without the consumer ever seeing the lost data, presenting a **phantom completion**: a partial response treated as a successful turn.

This is the dominant root of \"keeper does nothing\" symptoms observed downstream in masc-mcp: the agent thinks the turn finished, advances to the next turn with empty context, and loops idle.

## Changes

- Add two variants to \`Types.sse_event\`:
  - \`SSEParseFailed { raw; reason }\`
  - \`SSEUnknownEventType { event_type; raw }\`
- \`parse_sse_event\` emits these instead of returning \`None\`.
- \`Complete_stream_acc.accumulate_event\` treats both as stream errors (sets \`sse_error\`) so \`finalize_stream_acc\` returns \`Error _\` and the cascade layer can route to a different provider. Raw chunks are truncated to 200 chars in the diagnostic to avoid log explosion.

## What this PR does NOT do

- Does **not** classify the new errors as retryable in the cascade error model. That's the next PR — once that lands, a malformed Anthropic chunk will trigger automatic Codex/Gemini rotation instead of failing the turn.
- Does **not** touch \`parse_openai_sse_chunk\`, which has its own silent-None paths. Same fix pattern, separate PR for review hygiene.
- Does **not** touch other silent-default sites flagged in the audit (\`api_zai\` missing field → \`Unknown \"\"\`, token-counting magic divisors, hardcoded stop-reason mapping). Those are queued.

## Test plan

- [x] \`parse_sse_event\`: unknown event type → \`SSEUnknownEventType\` (regression test)
- [x] \`parse_sse_event\`: malformed JSON → \`SSEParseFailed\` with reason
- [x] \`parse_sse_event\`: missing type field → variant emitted (not silent \`None\`)
- [x] \`parse_sse_event\`: known event still parses (happy-path regression guard)
- [x] \`accumulate_event\`: \`SSEParseFailed\` marks \`sse_error\` with reason prefix
- [x] \`accumulate_event\`: \`SSEUnknownEventType\` marks \`sse_error\` with event_type
- [x] \`accumulate_event\`: large raw chunks truncated in diagnostic
- [x] \`dune build @lib/check\` clean
- [x] \`dune runtest lib/llm_provider/\` 14/14 pass

## Risk

Callers that pattern-match exhaustively on \`sse_event\` get a warning on the two new variants. Audited in-tree consumers (\`Complete_stream_acc\`, \`Streaming\`, \`Complete\`); only the accumulator matched exhaustively and is updated. Other sites use \`Some/None\` wrapping and pass the variant through unchanged.

External consumers of \`agent_sdk\` that match on \`sse_event\` exhaustively will see new variants — they should be added explicitly. The semantic is \"stream is corrupted, treat as error\" — recommend forwarding to the same handler as \`SSEError\`.


jeong-sik:
Reviewed Copilot's three threads — addressing partially in d3be4e5b, the rest deferred for design follow-up.

**Thread 1 — `lib/streaming.ml` public API still silent (fixed in d3be4e5b)**

You're right that the wildcard `_ -> ()` in `lib/streaming.ml::accumulate_event` (lines 89-90 before the patch) leaks the silent-drop bug into the stable `Agent_sdk.Streaming.create_message_stream` path. This commit applies the same handling that `Complete_stream_acc` already has (record `sse_error` with `sse_parse_failed:` / `sse_unknown_event_type:` prefix and a 200-char chunk preview) and replaces the wildcard with `MessageStop | Ping | ContentBlockStop _ -> ()` so a future `sse_event` constructor triggers a partial-match warning instead of silently regressing.

**Thread 2 — `event_type = ""` collapses two cases (acknowledged, deferring)**

Confirmed: `Cli_common_json.member_str "type"` returns `""` for both *missing* and *non-string* discriminator fields, so `SSEUnknownEventType { event_type = "" }` cannot be distinguished from `SSEUnknownEventType { event_type = "real_unknown_v3" }`. The right fix is probably either (a) a separate `SSEMissingEventType { raw }` variant or (b) tagging the parser path with `present : bool`. Option (a) cleanly preserves cascade-routing decisions but adds yet another constructor to a Stable type, which interacts with thread 3.

**Thread 3 — adding constructors to Stable `sse_event` is source-breaking (acknowledged, deferring)**

Also confirmed: `Agent_sdk.Types` re-exports from `lib/base/types.mli` (Stable). Adding constructors will break exhaustive matches in downstream callers on a minor-version bump.

Two paths I'm weighing:
- Internal-only variants: keep `sse_event` Stable as-is, route the new failure signals through a separate channel (e.g. `sse_error: string option ref` with a structured prefix, which is roughly what the accumulator already does today). External callers continue to pattern-match on the original constructors; the silent-drop fix is invisible to them but observable in the cascade error.
- Major-version intent: declare `sse_event` constructors public-extensible by documentation, ship in a major bump.

I'd prefer the first — the failure surface is downstream of pattern match anyway. But this needs a deliberate API decision rather than my own call.

Leaving Draft until threads 2/3 are decided.

PR #1383, opened 2026-05-05
feat(runtime): add durable relay delivery primitive

### LINKED ISSUE #1007: [mirror] harden OAS boundary relay truth in transport health
Mirrored from jeong-sik/masc-mcp#8365 — OAS boundary (this logic belongs in agent_sdk, not MASC).

## Problem

The OAS/MASC boundary now emits relay retry/drop metrics and lifecycle dispatch rejection counters, but two gaps remained:

- the transport dashboard did not surface the new relay health signals, so the UI could still look steady while the relay retry queue was backing up
- the OAS SSE bridge retry path could re-append the same durable event if broadcast failed after append succeeded

## Desired outcome

- transport health should show the OAS relay queue, retry, and drop state directly in the SSE section
- summary queue pressure should treat a non-empty relay retry queue as degraded
- agent-pool health should surface lifecycle dispatch rejection totals
- relay retry wiring should preserve the durable-append invariant: append once, retry broadcast without duplicating the stored event

## Validation

- targeted OCaml tests for `OAS Integration` and `Transport_metrics`
- targeted dashboard schema/component tests for `transport-health`

---
backlink: jeong-sik/masc-mcp#8365
auto-triage-tick: loop-30m-tick-1

jeong-sik:
## Summary

- Add `Relay_delivery`, a neutral two-stage relay primitive for adapters that persist an event before publishing it to a live transport.
- Preserve the durable-append invariant: if publish fails after persistence succeeds, retrying the returned pending value skips persistence and only retries publish.
- Track queue depth, retry totals, drop totals, and expose a `Runtime_health` probe so downstream transport-health surfaces can report relay pressure without scraping private state.

## Root Cause

The OAS boundary exposed event and durable journal primitives, but did not provide a reusable retry state machine for the common `persist -> publish` relay shape. Downstream bridges had to own both retry bookkeeping and durable append idempotence, which is exactly where duplicate append regressions can appear after a publish failure.

This keeps OAS ownership narrow: OAS provides the domain-neutral relay invariant and stats; downstream systems still own actual persistence, SSE/WebSocket transport, and dashboard schema.

## Validation

- `git diff --check`
- `opam exec -- dune exec --root=. ./test/test_relay_delivery.exe`

Fixes #1007

### DIFF
--- a/lib/agent_sdk.ml
+++ b/lib/agent_sdk.ml
@@ -133,6 +133,7 @@ module Tool_use_recovery = Tool_use_recovery
 module Tool_selector = Tool_selector
 module Lesson_memory = Lesson_memory
 module Event_forward = Event_forward
+module Relay_delivery = Relay_delivery
 module Metrics = Metrics
 module Progressive_tools = Progressive_tools
 module Autonomy_trace_analyzer = Autonomy_trace_analyzer
--- a/lib/agent_sdk.mli
+++ b/lib/agent_sdk.mli
@@ -102,6 +102,7 @@ module Autonomy_diff_guard = Autonomy_diff_guard
 module Metric_contract = Metric_contract
 module Lesson_memory = Lesson_memory
 module Event_forward = Event_forward
+module Relay_delivery = Relay_delivery
 module Metrics = Metrics
 module Progressive_tools = Progressive_tools
 module Autonomy_trace_analyzer = Autonomy_trace_analyzer
--- a/lib/relay_delivery.ml
+++ b/lib/relay_delivery.ml
@@ -129,9 +129,7 @@ let pending t = t.pending
 
 let stats t =
   let retry_total = t.retry_persist_total + t.retry_publish_total in
-  let drop_total =
-    t.drop_persist_total + t.drop_publish_total + t.drop_queue_total
-  in
+  let drop_total = t.drop_persist_total + t.drop_publish_total + t.drop_queue_total in
   { queue_depth = List.length t.pending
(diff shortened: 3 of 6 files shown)


jeong-sik:
## Review audit (auto-stamped)

Copilot reviewer threads scanned in batch. Reliability concern flagged for hand-review:

1. **CI is "skipping" not "running"** — Build/Test/Lint/OCaml-Structure jobs all show `skipping` due to `Detect Changed Surfaces` gate; only `CI Gate` (7s) actually executes. So `CI=SUCCESS` ≠ verified type-check pass. Copilot's typecheck claims (e.g. `int`/`float` mismatch, missing pattern branches) are NOT contradicted by green CI.

2. **At least one Copilot finding is hallucinated**: in #13081 thread on `caller_key`, Copilot listed `Graphql | Http | Startup | Auto_responder | Build_identity | Voice | Coord_identity | Dashboard | Http_routes | Test` as "newly added variants" requiring branches — none of these variants exist in the current `caller` type definition (verified against `lib/config/env_config_exec_timeout.mli` lines 8-38). Copilot was reviewing against an imagined/different diff.

3. **Recommendation**: Author hand-review each Copilot thread before mass-resolve. Real type errors cannot be distinguished from hallucinated ones via CI green alone.

Deferring batch resolution pending author judgement. Same audit cross-posted on sibling PRs where Copilot thread count ≥3 (#13067, #13081, #13085, #13093).

— Auto-posted by Claude Opus 4.7 (1M)

PR #1385, opened 2026-05-05
fix: wrap agent_sdk base library

jeong-sik:
## Summary

- Wrap `agent_sdk.base` so base modules no longer install global units like `Types`
- Re-export `agent_sdk.base` from the top-level `agent_sdk` library so `Agent_sdk.Types` remains usable downstream

## Why

`masc-mcp` also has an unwrapped `Types` unit. The previous `agent_sdk.base` packaging installed a second global `Types.cmi`, which made downstream builds fail with inconsistent assumptions over interface `Types`.

## Validation

- `env MASC_DUNE_THROTTLE=0 DUNE_LOCAL_JOBS=1 scripts/dune-local.sh build @install`
- `ocamlobjinfo` confirms `agent_sdk.cmi` imports `Agent_sdk_base__Types` instead of global `Types`

### DIFF
--- a/lib/agent_sdk.ml
+++ b/lib/agent_sdk.ml
@@ -38,6 +38,7 @@
 (** Re-export all modules (dependency-safe order) *)
 module Sdk_version = Sdk_version
 
+module Result_syntax = Result_syntax
 module Types = Types
 module Model_registry = Model_registry
 module Uncertain = Uncertain
--- a/lib/agent_sdk.mli
+++ b/lib/agent_sdk.mli
@@ -11,6 +11,7 @@
 (** {1 Core Modules} *)
 
 module Sdk_version = Sdk_version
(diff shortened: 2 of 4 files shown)


jeong-sik:
Re BC concern (lib/base/dune wrapped:true): tracking as deliberate API tightening — the change forces //etc. as the canonical names instead of unwrapped flat sibling modules. Downstream consumers of `agent_sdk` (top-level lib) are unaffected: `agent_sdk` re-exports `Types`, `Error`, `Context`, `Tool`, and now `Result_syntax` (commit ec5fde5b). Direct consumers of `agent_sdk.base` need to migrate to either `agent_sdk` top-level or qualified `Agent_sdk_base.*` paths; that is the intended outcome of the wrapping flip.

PR #1631, opened 2026-05-17
fix: add approval-required fail-closed policy

jeong-sik:
Summary:
- Add an explicit missing_approval_callback_policy to OAS hooks/options/builders.
- Preserve legacy execute-without-callback behavior by default while enabling Reject_without_callback fail-closed mode.
- Return deterministic non-retryable tool errors when ApprovalRequired has no approval callback under reject policy.

Verification:
- scripts/dune-local.sh build test/test_approval.exe test/test_builder.exe test/test_builder_coverage.exe
- ./_build/default/test/test_approval.exe (17 tests)
- ./_build/default/test/test_builder.exe (44 tests)
- ./_build/default/test/test_builder_coverage.exe (17 tests)
- git diff --check origin/main...HEAD

### DIFF
--- a/docs/rfc/RFC-OAS-021-approval-required-missing-callback-policy.md
+++ b/docs/rfc/RFC-OAS-021-approval-required-missing-callback-policy.md
@@ -1,0 +1,161 @@
+# RFC-OAS-021: ApprovalRequired Missing Callback Policy
+
+| | |
+|---|---|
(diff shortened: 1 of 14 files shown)


jeong-sik:
Closing duplicate: this same head SHA was already represented by merged PR #1630.

PR #2159, opened 2026-06-19
fix(typed_tool): improve result safety in lib/typed_tool.ml

jeong-sik:
P2 result modernization for lib/typed_tool.ml.

Changes:
- Add open Result_syntax.
- Flatten execute_parsed with let* instead of manual match/Ok/Error.

Skipped:
- check_descriptor still raises invalid_arg because create/create_with_context are public APIs; converting it would change their .mli signatures (breaking change).

Validation:
- ocamlformat --check lib/typed_tool.ml: passed
- MASC_DUNE_THROTTLE=0 ./scripts/dune-local.sh build lib/typed_tool.ml: passed
- dune exec test/test_typed_tool.exe: 16/16 passed
- dune exec test/test_typed_tool_safe.exe: 11/11 passed

### DIFF
--- a/lib/typed_tool.ml
+++ b/lib/typed_tool.ml
@@ -2,6 +2,8 @@
 
     @since 0.120.0 *)
 
+open Result_syntax
+
 (** Internal handler representation. Parametric variant — the type parameters
     ['input] and ['output] are threaded through both constructors. *)
 type ('input, 'output) handler_kind =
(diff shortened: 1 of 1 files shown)


jeong-sik:
### 적대 리뷰 · head f4d4c7bb — behavior-identical ✓

`execute_parsed`의 `match tool.parse json with Error e -> Error e | Ok input -> ...`를 `let* input = tool.parse json in Ok (input, run_handler ...)`로 정리. `let*`(= `Result.bind`)가 `Error e -> Error e` 단락을 정확히 대체 — error 의미 변화 0, behavior-identical. sound.

(근거: diff head f4d4c7bb 직접 정독. 차등 적대 리뷰 2026-06-19.)

PR #2176, opened 2026-06-26
fix(http): harden client error lifecycle

jeong-sik:
## Summary
- remove message-substring network classification from http_client and classify Unix/Eio errors through typed constructors
- keep Sys_error/Failure as Unknown instead of guessing from text
- make response drain failures explicit typed results with warning logs, while unknown exceptions propagate
- stop discarding the one-shot caller switch path; the callback receives the caller switch and one-shot transports are closed immediately after the request

## Verification
- scripts/dune-local.sh build test/test_http_client_cache.exe
- scripts/dune-local.sh runtest lib
- git diff --check

## Scope
This is the first HTTP client lifecycle hardening slice. It intentionally does not address retry.ml body-string classification or broader pipeline/env SSOT work.

### DIFF
--- a/lib/llm_provider/http_client.ml
+++ b/lib/llm_provider/http_client.ml
@@ -1496,6 +1496,12 @@ let%test "classify_unix_error: EHOSTUNREACH is Dns_failure" =
   classify_unix_error Unix.EHOSTUNREACH = Dns_failure
 ;;
 
+(* ── drain_response_body tests ───────────────────────── *)
+
+let%test "drain_response_body: complete source reports complete" =
+  Result.is_ok (drain_response_body (Eio.Flow.string_source "abc"))
+;;
+
 (* ── is_local_resource_exhaustion tests ──────────────── *)
 
 let%test "resource exhaustion: EADDRNOTAVAIL via Eio" =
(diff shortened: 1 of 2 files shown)


jeong-sik:
Rebased this PR onto current `origin/main` (`f6986af7`) and force-pushed existing branch `codex/oas-http-client-lifecycle-20260626` as `79298f2a3`.

What changed in the rebase:
- Kept the `#2184` typed HTTP classifier semantics from main. I skipped the stale classifier/drain commits from this branch instead of reintroducing the older `Multiple_io` / opaque reset behavior.
- Retained the useful lifecycle delta: the no-cache `with_client` path now uses `make_client` + `Fun.protect` and explicitly closes the one-shot transport after the callback returns.
- Updated the `get_sync` / `post_sync` docs to state that uncached connections close immediately after request completion.
- Kept only a small `drain_response_body` complete-source test from the lifecycle test cleanup.

Local focused verification:
- `ocamlformat --check lib/llm_provider/http_client.ml lib/llm_provider/http_client.mli`
- `git diff --check origin/main...HEAD`
- conflict-marker scan for `http_client.ml` / `http_client.mli`
- `scripts/dune-local.sh build @lib/llm_provider/runtest`

Current GitHub state after push: `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` because draft/checks are pending. CI is now the build authority for this rebased head.

PR #2182, opened 2026-06-26
fix(pipeline): remove dead error tagging wrapper

jeong-sik:
## Summary
- remove `Pipeline.tag_error`, which converted SDK errors to Error_domain context and then discarded the result
- wire pipeline stages directly so error flow is explicit and allocation-free
- remove identity-only inline tests for the deleted wrapper and rename Error_domain tests to their actual scope

## Review Notes
The removed wrapper did not publish, log, return, or attach the computed stage context. Keeping it made the pipeline look more observable than it was. This PR keeps existing `sdk_error` propagation behavior and removes only the dead diagnostic computation.

## Verification
- `ocamlformat --check lib/pipeline/pipeline.ml test/test_pipeline_deep.ml test/test_pipeline.ml`
- `git diff --check`
- residue scan: no `tag_error` references remain in `lib/pipeline/pipeline.ml`
- Dune build/test intentionally left to GitHub CI per hardening review workflow.

### DIFF
--- a/lib/pipeline/pipeline.ml
+++ b/lib/pipeline/pipeline.ml
@@ -751,17 +751,6 @@ let emergency_compact ?raw_trace_run agent ?limit () =
 
 (* ── Pipeline coordinator ────────────────────────────────── *)
 
-let tag_error stage result =
-  match result with
-  | Ok _ as ok -> ok
-  | Error e ->
-    let poly = Error_domain.of_sdk_error e in
-    let _ctx = Error_domain.with_stage stage poly in
-    (* Stage context is created for diagnostics;
-       we still propagate sdk_error for backward compat *)
-    Error e
-;;
-
 let run_turn ~sw ?clock ~api_strategy ?raw_trace_run agent =
   (* Stage 1: Input *)
   let* () =
@@ -774,7 +763,7 @@ let run_turn ~sw ?clock ~api_strategy ?raw_trace_run agent =
       ; extra = []
       ; links = []
       }
-      (fun _tracer -> stage_input ?raw_trace_run agent |> tag_error "input")
+      (fun _tracer -> stage_input ?raw_trace_run agent)
   in
   (* Stage 2: Parse *)
   let prep, original_config, turn_params =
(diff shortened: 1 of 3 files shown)


jeong-sik:
# Review: PR #2182 — fix(pipeline): remove dead error tagging wrapper

- **State**: OPEN, draft
- **CI**: ✅ green — `Build & Test (OCaml 5.4.1)`, `ratchet`, `Boundary Lint`, `Draft Auto-Merge Guard` all passed.
- **Files changed**: `lib/pipeline/pipeline.ml`, `test/test_pipeline.ml`, `test/test_pipeline_deep.ml`
- **Net change**: -70 LoC

## Summary
Removes the identity `tag_error` wrapper from `Pipeline.run_turn`, wires `stage_input` / `stage_route` / `stage_collect` / `stage_output` directly, deletes tautological inline tests, and renames the remaining `Error_domain` tests so they exercise the conversion helpers independently of the coordinator.

## Findings

### P1 — Merge-order hazard with #2174
`#2174` (`adversarial-review-oas-fixes`) modifies the same `tag_error` function and all four call sites. It upgrades the wrapper to log the computed `Error_domain` context and adds `?clock` to the stage calls:

- `#2174` diff touches `tag_error` definition around `lib/pipeline/pipeline.ml:1647` and preserves `|> tag_error "..."` at lines ~1663, ~1743, ~1787, ~1791.

This PR deletes that wrapper entirely. Whichever lands second will hit non-trivial conflicts and may accidentally resurrect dead code or drop `#2174`’s `?clock`/logging work.

**Fix**: Rebase this branch on top of `#2174` once it merges, or ask `#2174` to drop its `tag_error` changes and absorb this cleanup first.

### P2 — `Error_domain` now has no internal production callers
After `tag_error` is gone, none of `Error_domain.of_sdk_error`, `with_stage`, `to_sdk_error`, `is_retryable`, etc. are used inside `lib/` — only in tests. If the module is intended as public SDK surface, add a brief `.mli` note; otherwise schedule its removal. This PR does not need to do it, but the cleanup makes the dead-API status visible.

### P2 — Tests still assert on serialized string prefixes
The renamed `test_stage_context_*` tests in `test/test_pipeline_deep.ml` still do manual prefix checks:

```ocaml
String.length s >= String.length prefix
&& String.sub s 0 (String.length prefix) = prefix
```

Use `String.starts_with ~prefix` instead. The rename is a good opportunity to clean this up.

## What is good
- Pure dead-code removal; observable pipeline behavior is unchanged.
- No new env vars, local paths, mutable state, or Eio lifetime/cancellation changes.
- Format and ratchet checks pass; CI is green.

## Cluster context
This PR is part of the current OAS pipeline/http hardening cluster: #2174, #2176, #2177, #2179, #2184. Its blast radius is small (one wrapper + tests) and safe **once the #2174 ordering is resolved**.

PR #2190, opened 2026-06-26
fix(oas): reset idle counter on non-tool-use turns and idle Skip

jeong-sik:
## Summary
Agent_turn.update_idle_detection only sees ToolUse blocks, so a run of repeated identical tool calls followed by a text/thinking/EndTurn/etc. turn would keep the stale consecutive_idle_turns counter. Reset it on every non-tool-use stop reason and on IdleSkipped so the idle loop detector starts fresh when the agent returns to tool use.

## Test Plan
- test/test_agent_turn.ml: 39 tests pass, including new "reset breaks idle streak" case.
- Full CI green required before merge.

### DIFF
--- a/lib/agent/agent_turn.ml
+++ b/lib/agent/agent_turn.ml
@@ -487,6 +487,10 @@ let update_idle_detection ~idle_state ~tool_uses =
     ~tool_uses
 ;;
 
+let reset_idle_detection () =
+  { new_state = { last_tool_calls = None; consecutive_idle_turns = 0 }; is_idle = false }
+;;
+
 (** Default per-tool-result character cap.
     Aligned with Claude Code's DEFAULT_MAX_RESULT_SIZE_CHARS (50,000).
     Results exceeding this are truncated with a marker at creation time,
--- a/lib/agent/agent_turn.mli
+++ b/lib/agent/agent_turn.mli
@@ -232,6 +232,9 @@ val update_idle_detection_with_normalizer
   -> tool_uses:Types.content_block list
   -> idle_result
 
+(** Reset idle detection state after a non-tool-use turn or an idle Skip. *)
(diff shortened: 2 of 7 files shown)


jeong-sik:
Thanks for the adversarial review.

**Addressed in this follow-up:**

1. **P1 — direct mutable-field assignment** — Resolved.
   - Added mutex-wrapped helpers `Agent_types.set_idle_state : t -> idle_state -> unit` and `Agent_types.reset_idle_state : t -> unit` in `lib/agent/agent_types.{ml,mli}`.
   - Replaced the direct assignments in `Pipeline.stage_execute` (idle update after tool-use detection) and `Pipeline.stage_output` (reset on `IdleSkipped`, non-tool-use stops, and `Unknown`) with calls to these helpers.
   - This removes the contract violation noted in `Agent_types.mli` and centralizes the two mutation sites.

2. **P2 — reset before `on_stop` hook** — Resolved.
   - For non-tool-use stop reasons (`EndTurn`, `MaxTokens`, `StopSequence`, `Refusal`, `PauseTurn`, `Compaction`, `ContextWindowExceeded`), the reset now happens *after* `invoke_hook_with_trace ~hook_name:"on_stop"`, so observers can read the actual `consecutive_idle_turns` value that drove the turn.
   - `IdleSkipped` and `Unknown` remain reset before returning because those paths invoke no observer hook in `stage_output`.

**Verification:**
- `dune test test/test_agent_turn.ml` — 39 tests pass.
- `dune test test/test_pipeline.ml` — 45 tests pass (including `output resets idle on end turn` and `output resets idle on unknown`).
- `dune test test/test_agent_race.ml` and `test/test_pipeline_common_coverage.ml` — pass.

No other code was touched. Ready for re-review.

PR #2192, opened 2026-06-27
refactor(env): consolidate int/float env parsing to Cli_common_env

jeong-sik:
## Summary

- Adds a canonical `float` env parser to `lib/llm_provider/cli_common_env.ml`/`mli`.
- Refactors `lib/defaults.ml` and `lib/base/util.ml` to delegate `int_env_or`/`float_env_or` to the canonical parser.
- Updates `lib/tool_result_store.ml` to use `Cli_common_env.get` for non-empty, trimmed lookups.
- Documents the behavior change that `OAS_MCP_OUTPUT_MAX_TOKENS=0` is now accepted as a valid non-negative value instead of falling back to `25_000`.

## Validation

- `dune build @all --root .` :white_check_mark:
- `dune runtest --root . --force` :white_check_mark: all suites pass


jeong-sik:
[Adv Review] oas#2192 — centralize env parsers and accept non-negative values

Head reviewed: `ce3f86b5ee4292899ccf5647ef03e806cb3a5da4`

Overall: the PR is a useful cleanup — it consolidates duplicated `int_env_or`/`float_env_or` implementations, adds the missing `float` variant to `Cli_common_env`, and correctly changes the semantics so that `0` is accepted instead of falling back to the default. However, the consolidation is incomplete and leaves a silent-failure hole. Do not merge until the P1 items below are resolved and CI is fully green.

---

### P1 — Incomplete SSOT / inconsistent diagnostic channel in `Defaults`

`Defaults.int_env_or` and `Defaults.float_env_or` now delegate to `Llm_provider.Cli_common_env`, which emits warnings via `Diag.warn` under the context `"cli_common_env"`:

```ocaml
(* lib/defaults.ml, new lines 16-17 *)
let int_env_or default var = Llm_provider.Cli_common_env.int ~default var
let float_env_or default var = Llm_provider.Cli_common_env.float ~default var
```

But `bool_env_or` still uses the local `warn_invalid_env` helper that writes to a module-specific `Log` sink:

```ocaml
(* lib/defaults.ml, new lines 19-29 *)
let bool_env_or default var =
  match Llm_provider.Cli_common_env.get var with
  | None -> default
  | Some v ->
    (match String.lowercase_ascii v with
     | "1" | "true" | "yes" | "on" -> true
     | "0" | "false" | "no" | "off" -> false
     | _ ->
       warn_invalid_env ~var ~raw:v ~expected:"boolean";
       default)
```

This means invalid env values for int/float are reported through `Diag`, while invalid booleans are reported through `Log`/`Defaults`. If the goal of this PR is a single source of truth for env parsing, the three parsers should share the same warning path. Either extend `Cli_common_env` with a `bool ~default` variant and route `bool_env_or` through it, or keep all three in `Defaults` and delete the `Cli_common_env` delegation. The current half-migrated state is worse than the pre-PR duplication because it makes observability/alerting depend on which env var is malformed.

---

### P1 — Silent acceptance of negative values in `Tool_result_store`

`Tool_result_store.int_of_env` was partially migrated to use `Cli_common_env.get`, but it still does its own parsing with `int_of_string_opt` and no lower-bound validation:

```ocaml
(* lib/tool_result_store.ml, new lines 24-28 *)
let int_of_env name =
  match Llm_provider.Cli_common_env.get name with
  | None -> None
  | Some s -> int_of_string_opt s
;;
```

A value like `OAS_TOOL_RESULT_THRESHOLD="-100000"` will be accepted silently and stored in `threshold_chars`. The same variable, if it went through `Defaults.int_env_or`, would be rejected with a warning. This undermines the PR’s stated consolidation and is a silent failure. Either:

1. Use `Llm_provider.Cli_common_env.int` here (with `allow_negative:false`) and drop the local `int_of_env` helper, or
2. Explicitly document and test that negative thresholds are intentionally permitted for these three variables.

Option 1 is strongly preferred if this PR is meant to centralize env parsing.

---

### P2 — Duplicated validation logic in `Cli_common_env`

`Cli_common_env.int` and the new `Cli_common_env.float` are copy-paste variants of the same pattern (trim, empty check, parse, negative guard, warning). Extract a small private helper parameterized over the parser, comparator, and format string to prevent the two implementations from drifting apart in future edits.

---

### P2 — Fragile test assertions

The new float tests rely on `String.contains msg '-'` and `'n'` to detect negative/non-numeric errors:

```ocaml
(* lib/llm_provider/cli_common_env.ml, new tests *)
&& List.exists
     (fun (level, ctx, msg) ->
        level = Diag.Warn && ctx = "cli_common_env" && String.contains msg '-')
     !warnings)
```

This is brittle (e.g. a future message rewording could remove the literal minus sign). Assert on a stable substring such as `"is negative"` or `"is not a float"`, or expose and match on a structured error value.

---

### P2 / informational — semantics change is visible and tested

The change from `v > 0` to `v >= 0` is clearly documented in `CHANGELOG.md`, `util.mli`, `defaults.mli`, and the updated tests in `test_defaults.ml`, `test_mcp_coverage.ml`, and `test_mcp_deep.ml`. That part is handled correctly.

---

### CI status

- `Build & Test (OCaml 5.4.1)` — `IN_PROGRESS` at review time.
- `ratchet`, `Boundary Lint (Core→CDAL)`, `Draft Auto-Merge Guard` — `SUCCESS`.
- `Lint`, `OCaml Format`, `Version Consistency`, `Transport Drift Gate`, `SDK Independence Gate`, `Release Tag Drift`, `TLA+ Model Checking` — `SKIPPED`.
- `mergeStateStatus`: `BLOCKED` (likely because the required build is still in progress and/or the PR is in draft).

Do not merge until `Build & Test (OCaml 5.4.1)` reports `SUCCESS`.

---

Bottom line: please route `bool_env_or` through the same `Cli_common_env` warning path, close the negative-value hole in `Tool_result_store`, and clean up the duplicated `int`/`float` validation before requesting another review.

PR #2208, opened 2026-06-28
fix(http): bind one-shot client cleanup to caller switch

jeong-sik:
## Summary
- bind one-shot HTTP client cleanup to the caller-provided Eio switch
- keep eager close-on-return with close-once so long-lived caller switches do not retain idle FDs
- remove the nested one-shot Eio.Switch.run path for synchronous requests without a cache

## Verification
- git diff --check
- opam exec -- ocamlformat --check lib/llm_provider/http_client.ml

Full build/test intentionally left to GitHub CI per repo workflow.

### DIFF
--- a/lib/llm_provider/http_client.ml
+++ b/lib/llm_provider/http_client.ml
@@ -3,11 +3,11 @@
     Wraps Eio + cohttp-eio with TLS. All network and HTTP-level errors
     are captured as {!http_error} so callers do not need [try/with].
 
-    Each synchronous request without a [connection_cache] runs inside
-    its own [Eio.Switch.run] scope so the underlying TCP connection and
-    its file descriptor are released as soon as the response body is
-    fully consumed. With a cache, connections are bound to the cache's
-    switch and reused until eviction or switch release.
+    Each synchronous request without a [connection_cache] registers its
+    transport cleanup with the caller's switch and also closes the
+    transport as soon as the response body is fully consumed. With a cache,
+    connections are bound to the cache's switch and reused until eviction
+    or switch release.
 
     @since 0.45.0 *)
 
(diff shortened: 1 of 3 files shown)


jeong-sik:
Current-head review on `d875183d9ecd691d4b10df5b83b8f4679396256a`: request changes.

The implementation shape is plausible: the one-shot path now registers `close` on the caller switch, wraps it with `close_once`, and still closes eagerly after `f` returns. But this is a production resource-lifetime change for the non-cached `get_sync` / `post_sync` path, and the PR changes only `lib/llm_provider/http_client.ml`. There is no regression test added for the exact invariant this is fixing: normal return closes exactly once after full body read, cancellation releases the transport through the caller switch, and long-lived caller switches do not retain idle FDs after repeated one-shot calls.

There is also a public-doc drift: `http_client.mli` still says the non-cached `get_sync` / `post_sync` connection is closed when `sw` is released, while the implementation now also closes it immediately after `f` returns. That difference is intentional per the `.ml` comment, so the public interface should say it explicitly.

Please add focused regression coverage or point to an existing test that exercises this non-cached `with_client` path under normal return + cancellation, and update the `.mli` lifecycle text. Until then, this is not enough proof for an FD/leak cleanup fix.

[근거] `git diff --name-only origin/main...origin/pr-2208` and `git show origin/pr-2208:lib/llm_provider/http_client.ml | nl -ba | sed -n '850,860p'` at 2026-06-28T12:20:37+09:00, confidence High: current diff is one file; implementation lines 850-859 change the one-shot lifetime; `http_client.mli` lines 229-256 still document release-on-switch only; `gh pr checks 2208 --json name,state,bucket,link` shows Build & Test still IN_PROGRESS.

jeong-sik:
Addressed the review blocker on current head `e6e0c052`.

Changes:
- added raw-loopback regression coverage for the one-shot no-cache `get_sync` lifecycle;
- the normal-response case now asserts the server observes client EOF before the parent switch is released;
- the timeout/error case asserts the same cleanup path after an `Http_operation` timeout;
- updated `Http_client.mli` so the public get/post contract says no-cache cleanup is both switch-bound and eager after body consumption.

Local light validation only, per CI-owned Dune policy:
- `ocamlformat --check lib/llm_provider/http_client.mli test/test_http_client_cache.ml`
- `git diff --check`

CI is attached on the pushed head and remains the build authority.

PR #2223, opened 2026-06-28
[codex] fix agent runtime hardening for keeper safety

jeong-sik:
## Summary

Hardens the OAS side of the Keeper P0/P1 audit:

- make execution timeout own the child switch and avoid cancellation-swallowing handoff paths
- centralize HTTP error to SDK error mapping
- deduplicate cache content-block serialization through the canonical API codec
- classify parser/protocol failures without retryable synthetic 5xx/network wrappers
- fail closed for unknown OpenAI-compatible model tool-choice capability
- fix slot scheduler cancellation/release invariants
- preload/catalog state through atomic snapshots instead of request-path lazy force
- require explicit runtime store roots and remove cwd/HOME fallback from embedded paths
- make tool selection fail closed instead of BM25 fallback on LLM classifier/reranker failure
- remove config fake-tool acceptance and require MCP connection errors to surface

## Why

MASC Keeper liveness depends on OAS not converting local parser bugs, path drift, cancellation, hook failures, or unsupported model/tool capabilities into silent fallback or retry noise.

## Validation

- `git diff --check origin/main...HEAD`
- `git diff --name-only origin/main...HEAD -- '*.ml' '*.mli' | xargs ocamlformat --check`
- static scans for hardcoded `/Users/...`, runtime-store cwd fallback, path helper fallbacks, prose tool-selector parsing, and BM25 fallback-on-failure language

Dune build is intentionally left to CI per operator instruction.


jeong-sik:
Current-head review for `ccb24fddd2b70d724e59fc9364e3349e7245f8f5`.

This is still blocked, and the Build & Test failure is pointing at a real runtime-contract regression, not just stale inline expectations.

CI evidence from `Build & Test (OCaml 5.4.1)`:

- `backend_openai.ml`: 8/93 inline tests fail, several by throwing `Yojson__Safe.Util.Type_error` while reading request fields that are now `null`.
- `capabilities.ml`: 22/46 known-model capability tests fail (`glm-*`, `nvidia-*`, `qwen*`, `gemma*`, etc.).
- `pricing.ml`: 16/99 pricing tests fail for known Gemini/GLM families.
- `discovery.ml`, `complete.ml`, `complete_sampling.ml`, and `provider_config.ml` also fail capability/provider behavior assertions.

The source shape explains the breadth: `lib/llm_provider/model_catalog.ml:174-203` now loads only `OAS_MODEL_CATALOG`, and `lib/llm_provider/capabilities.ml:758-770` returns `None` when neither a runtime override nor that env-loaded catalog exists. But the rest of the SDK still exercises known-model behavior in ordinary test/runtime paths. So an unset catalog no longer just removes an ambient fallback; it degrades known-model capabilities/pricing and lets request building continue with missing/null fields.

That is not fail-closed enough for this PR's stated safety goal. Either every binary/test/embedding path must install or preload an explicit catalog before any provider request/capability/pricing path can run, or the SDK needs a packaged non-ambient catalog source. If the intended contract is "no catalog means no known model metadata", then request construction/pricing should fail with a clear config error before serialization, not silently return `None`/zero/omitted fields and then crash or emit wrong provider payloads.

Please keep MASC #22625 semantically blocked on this until OAS is green and released/pinned; otherwise the MASC side can look hardened while consuming an OAS runtime that no longer resolves the model metadata it depends on.

PR #2304, opened 2026-06-30
fix(llm_provider): close #2236 CoT re-injection loop, provider-agnostic

jeong-sik:
## Summary

Closes the #2236 CoT re-injection / parrot loop and retires the dead
`reasoning_visibility` vocabulary, then proves the fix is uniform across every
provider with a deterministic, mutation-tested harness.

**Root cause.** OpenAI-compatible response parsing *promoted* model reasoning (a
`Thinking` content block) into a `Types.Text` answer block. That erased the type
distinction between reasoning and answer, so on the next turn the request
serializer re-fed the reasoning as the assistant's `content` (answer) and the
model re-reasoned over it — the parrot / infinite-recursion loop.

**Fix.** Keep reasoning typed as `Thinking` end-to-end. Reasoning-only replies
still reach the user through the display path (`Api.text_blocks_to_string`
includes `Thinking` blocks). Multi-turn replay follows each provider's
`reasoning_dialect` policy: tool-call turns carry the reasoning in the dedicated
`reasoning_content` wire field (satisfying the DeepSeek / MiMo "400 if tool-call
reasoning is missing" contract), plain answer turns drop it. Reasoning never
re-enters the `content` channel.

**Why this is provider-agnostic.** The thinking wire contract is a property of
the *serving platform*, not the model: the same DeepSeek weights expose
`message.thinking` + a `think:` request param on Ollama's `/api/chat`
(`backend_ollama.ml:269`), `reasoning_content` on the native DeepSeek / OpenAI
reasoning endpoint (`backend_openai_parse.ml:292`), and `reasoning` on Ollama's
OpenAI-compat `/v1` (`:294`) — at least four distinct wire conventions. OAS
derives the replay dialect from *capabilities / backend* (the platform), not from
a model-name string, so it never conflates "DeepSeek on Ollama" with "DeepSeek
native". Crucially, all four wire shapes converge to one internal type
(`Thinking { … }`); the per-platform divergence is isolated to the backend
parser, and everything above it is unified. That convergence is what makes the
anti-parrot guarantee hold for any provider — and this PR keeps that convergence
intact by ensuring `Thinking` is never promoted to a `Text` answer block.

The structural guarantee covers any provider whose reasoning arrives in one of
the recognized wire fields above. A genuinely novel field name (a hypothetical
new platform that emits neither `thinking`/`reasoning_content`/`reasoning` nor an
inline `<think>` tag) still needs a one-line parse-side mapping into `Thinking` —
the irreducible per-platform wire work — but inherits the replay / anti-parrot
contract for free once mapped.

## What changed

- **Phase 1 (`b9bf91a07`)** — remove the reasoning→`Text` promotion from both the
  non-streaming parser and the streaming accumulator; reasoning stays `Thinking`.
- **Phase 2 (`e5fbf63ee`)** — retire the `reasoning_visibility` capability
  vocabulary (19 files, −204 net lines). Its only runtime effect was the
  promotion above; every other use parsed and stored a value that was never read.
  Removed from the public API: `Capabilities.reasoning_visibility_override`
  (type + record field + the four `Force_*`/`Default_*` constructors), the same
  `Provider` re-exports, and `Reasoning_dialect`'s `visibility` field /
  `visibility_to_string`. `ollama_cloud_capabilities` (whose only difference was
  `Force_visible_text`) becomes an explicit alias of `ollama_capabilities`. The
  `reasoning_visibility` capability-manifest key is now rejected fail-closed as
  an unknown field rather than silently ignored. Schema + docs updated to match.
- **`d7e095c29`** — strengthen the regression guard (assert a reasoning-only reply
  serializes to *empty* content, not merely "not equal to the reasoning"); add
  CHANGELOG entries documenting the fix and the API removal.
- **`942f3180b`** — provider-agnostic anti-parrot harness.

## Why retire the vocabulary instead of keeping it

A dead config field that silently no-ops is a Silent-Failure footgun: future
code learns it as a precedent and AI-generated catalog entries keep setting it.
Removing it makes the retirement a compile-time event (the OCaml exhaustiveness
checker flagged all 13 consumer sites) and a fail-closed manifest rejection. No
live catalog uses the key (verified). No behavior change to thinking toggle,
replay, streaming, or effort.

## Verification

Provider-agnostic harness `test/test_provider_agnostic_reasoning_replay.ml`
drives all 11 exported provider capability profiles (anthropic, kimi, gemini,
ollama, ollama_cloud, dashscope, glm, openai_compat ×2, provider_l, default)
through the live serializer `dialect_messages_of_message` and asserts:

- **INV1 (anti-parrot):** reasoning text never appears in the answer `content`
  field — every provider × reasoning_only / answer / tool_call shapes.
- **INV2 (uniform interleaving):** the wire matches
  `Reasoning_dialect.should_replay_reasoning` exactly — a tool-call turn carries
  `reasoning_content` iff the policy replays; a plain answer turn drops it.
- **INV3 (recursion closure):** replaying a reasoning-only reply 5 rounds never
  accumulates into `content`.
- **End-to-end:** `parse(reasoning_content response) → serialize` round-trip
  keeps reasoning typed as `Thinking` and out of `content` for every provider.

**Mutation-tested:** temporarily reintroducing the promotion makes the
round-trip case go RED (exit 1) on both the parse-side guard and the codec
guard; reverting restores GREEN. The GREEN result has teeth.

Local (with `OAS_MODEL_CATALOG=models.toml`): `dune @check` green; ocamlformat
clean; test suites green —
`test_provider_agnostic_reasoning_replay` (4),
`test_backend_openai_codec` (33),
`test_stream_accumulator` (27),
`test_thinking_control_dialects` (29),
`test_capabilities` (61, incl. the fail-closed retirement guard),
`test_provider_registry` (43).

## Self-review (Best Programmer)

- **Goal:** close the #2236 parrot loop and prove provider-agnostic uniform
  interleaving / tool-use / CoT.
- **Changed files:** parser + streaming accumulator + serializer plumbing; the
  `reasoning_visibility` removal across capabilities / dialect / 3 catalog
  parsers / provider projection / manifest schema + docs; CHANGELOG; 1 new test
  file + tightened codec test.
- **Local validation:** above. RED→GREEN mutation proof for the loop closure.
- **Remaining unverified:** the final pre-answer reasoning of a tool-using
  DeepSeek round (Thinking 1.3) is dropped on the next user turn; the DeepSeek
  diagram and the doc text disagree on this edge. OAS's behavior is 400-safe and
  the safe direction (dropping reasoning cannot cause the loop). Worth a live-API
  confirmation as a follow-up, not a blocker for this change.

## Scope note

A unified "Adaptive Thinking + Effort" control surface (one effort/adaptive knob
mapped to every provider's native field) is a separate forward enhancement on
top of the existing `Reasoning_effort`, tracked separately — not required for the
loop closure or the uniform-interleaving guarantee this PR proves.


jeong-sik (commented):
## 적대 리뷰: #2304 (CoT re-injection loop fix)

**verdict: LGTM (verified-correct, HONEST POSITIVE)**

### 강점 (3개)

1. **Root cause 정확성**: Phase 1의 `Visible_text` promotion이 reasoning을 `Text`로 변환 → 다음 턴 serializer가 `content`로 재투입 → parrot loop. promotion 자체가 root cause. **올바른 fix 위치 (parse-time mutation 안 함)**.

2. **Phase 2 retire = OCaml idiom 정답**: 죽은 config field의 silent no-op = Silent-Failure footgun. 13 consumer sites compile-time flag는 OCaml exhaustiveness checker의 가장 강한 보증. "노 Silent Failure" 원칙 정확 부합. fail-closed unknown field rejection = catalog hygiene까지 보강.

3. **Mutation test = GREEN에 이빨**: 잠시 promotion 재도입 → RED, revert → GREEN. 1-cycle revert로 fix 검증 = self-review에서 "GREEN result has teeth" 진술 정확.

### 검증 (PR body 진술 cross-check)

- `Api_common.text_blocks_to_string` (`lib/llm_provider/api_common.ml:35-44`)가 `Thinking { content; _ }`를 `Some`으로 매핑 → display 경로에 reasoning 포함. **"Reasoning-only replies still reach the user through the display path" 진술 검증 완료**.
- `RedactedThinking`만 `None`으로 제외 = 의도된 동작.

### Self-review 정직성 인정

"the DeepSeek diagram and the doc text disagree on this edge. OAS's behavior is 400-safe and the safe direction (dropping reasoning cannot cause the loop). Worth a live-API confirmation as a follow-up, not a blocker for this change" — **이 disclosure가 정확**. 400-safe 방향은 parrot loop 방지 + cross-provider contract 보존 모두 만족.

### Minor (P3, blocker 아님)

- `ollama_cloud_capabilities`가 `ollama_capabilities`의 explicit alias가 되면서 capability 차별점 사라짐. live catalog 0건이므로 안전. 차후 새 capability 차별 필요 시 `ollama_cloud_capabilities`를 다시 분리해야 함.
- 후속 follow-up 후보: DeepSeek `Thinking 1.3` final pre-answer reasoning drop edge의 live-API confirmation.

### Cross-repo coupling 점검

- MASC `keeper_chat_oas_stream_bridge.ml:72-91`이 `TextDelta→Text_delta` / `ThinkingDelta→Oas_thinking_delta` / `SignatureDelta→byte count`로 typed 분리. **#2304 promotion 제거 후에도 bridge typed boundary 정확** (string-match 없음).
- Display dedup은 ts-scoped → cross-turn 진짜반복 충실표시 = 올바름.

### 결론

- 라이브 root 직접 fix + retirement + mutation test = 3중 방어선
- 4 wire shapes (`message.thinking`/`reasoning_content`/`reasoning`/DeepSeek native) 모두 `Thinking { … }`로 수렴 = provider-agnostic 보장
- PR body self-review 정직 = 신뢰

CI green 후 Ready 전환 권고.

PR #2309, opened 2026-06-30
fix(glm): route thinking fields through shared builder

jeong-sik:
## Summary
- Route native GLM detection in the OpenAI-compatible request builder through the typed Provider_config.is_zai_glm_config predicate, so kind=Glm and ZAI OpenAI-compatible GLM use the same canonical thinking-field path.
- Remove Backend_glm's duplicate manual thinking object emitter; Backend_glm now only keeps GLM-specific post-processing for tool_choice normalization and tool_stream.
- Add a regression guard that fails if Backend_glm starts emitting thinking-control wire fields directly before its inline tests section.

## Verification
- git diff --check
- Dune build/test: intentionally left to GitHub CI per maintainer workflow.

## RFC-OAS-029
Closes part of S2.1/S9.2 by reducing GLM thinking field construction to the shared Reasoning_dialect.request_control_fields path instead of a second backend-local builder.

### DIFF
--- a/lib/llm_provider/backend_glm.ml
+++ b/lib/llm_provider/backend_glm.ml
@@ -1,7 +1,7 @@
 (** ZhipuAI Glm native backend.
 
     Openai wire format with Glm-specific extensions:
-    - [thinking] parameter: [{"type":"enabled","clear_thinking":true}]
+    - thinking control parameter with type=enabled and clear_thinking=true
     - [reasoning_content] in response message and streaming delta
     - String error codes (e.g., ["1305"])
     - Temperature range 0-1 (not 0-2)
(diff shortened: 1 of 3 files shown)


jeong-sik:
**반경 확인 (연계 PR #2314)**: #2314 "read reasoning deltas from dialect"가 `Reasoning_dialect`를 읽지만 소비하는 건 `.streaming`(`Delta_field`/`No_streaming_reasoning`/`Template_parser`)뿐이고 `.visibility`는 읽지 않습니다. dialect를 읽는 연계 PR 맥락에서도 `visibility` 소비처는 0으로 확인됩니다. axis별 마이그레이션(#2314=streaming)이라면 visibility는 아직 없는 후속 PR용 forward-prep인데, 그게 의도면 P1.2의 "구체적 소비 지점"에 그 후속 작업을 적어주시면 #2304 retirement 근거와 구분됩니다.

PR #2444, opened 2026-07-02
fix(llm_provider): harden redaction for large media payloads

jeong-sik:
## Summary
- make secret prefix checks allocation-free instead of using String.sub for every position
- avoid per-character full-tail rescans when a large payload contains no redaction prefixes
- collapse base64 media data URLs before token scanning so trace/log persistence does not replay full inline image payloads
- add large non-secret payload and media data URL regression tests for image/Fusion-sized strings

## Incident evidence
- MASC 8935 health timed out while main_eio.exe stayed running
- sample of PID 89985 showed the hot path in Llm_provider__Secret_redactor plus caml_blit_bytes, with 1.6G physical footprint / 2.3G peak
- screenshots showed dashboard transport offline / client WS disconnected / logs stuck loading

## Local MASC recovery proof
- locally pinned agent_sdk to this OAS worktree with AGENT_SDK_PIN_URL and rebuilt MASC using MASC_SKIP_PIN_CHECK=1 because the repo pin still intentionally points at the old OAS main SHA
- replaced wedged PID 89985, then refreshed the listener after the media data-URL hardening; current rebuilt listener PID is 92896 on 127.0.0.1:8935
- GET /health?full=1 returned 200 in 0.546029s after the media hardening rebuild, and later 200 in 0.001469s
- GET /dashboard#keepers returned 200 in 0.581833s
- GET /api/v1/dashboard/logs?limit=50 returned 200 in 0.549899s, then 200 in 0.000990s
- sample of PID 92896 did not show the previous Secret_redactor/String.sub hot path; main thread was in poll

## Validation
- ocamlformat --check lib/llm_provider/secret_redactor.ml lib/llm_provider/secret_redactor.mli
- git diff --check -- lib/llm_provider/secret_redactor.ml lib/llm_provider/secret_redactor.mli
- bash scripts/hardening-ratchet.sh --check
- bash scripts/ci-code-smell-ratchet.sh --check
- scripts/dune-local.sh build @lib/llm_provider/runtest

### DIFF
--- a/lib/llm_provider/secret_redactor.ml
+++ b/lib/llm_provider/secret_redactor.ml
@@ -111,8 +111,15 @@ let find_data_url_comma s pos =
     else (
       match s.[i] with
       | ',' -> Some i
-      | ch when is_data_url_header_terminal ch -> None
-      | _ -> loop (i + 1))
+      | '"' | '\'' | '<' | '>' | ' ' | '\t' | '\n' | '\r' -> None
+      | '\000' .. '\b'
+      | '\011' | '\012' | '\014'
+      | '\015' .. '!'
+      | '#' .. '&'
+      | '(' .. '+'
+      | '-' .. ';'
+      | '='
+      | '?' .. '\255' -> loop (i + 1))
   in
   loop pos
 ;;
@@ -345,9 +352,8 @@ let%test "redact_string masks private key block" =
 ;;
 
 let%test "redact_json preserves structure" =
-  match redact_json (`Assoc [ "key", `String "Bearer tok" ]) with
-  | `Assoc [ ("key", `String "Bearer [REDACTED]") ] -> true
-  | _ -> false
+  redact_json (`Assoc [ "key", `String "Bearer tok" ])
+  = `Assoc [ "key", `String "Bearer [REDACTED]" ]
 ;;
 
 let%test "redact_string collapses base64 media data url" =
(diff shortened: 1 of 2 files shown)


jeong-sik:
Current-head follow-up @ 524dccdf8ac91b4d258d2c26ca2a0f3e024a4548.

P0: Patched a redaction bypass in the media-collapse path. The PR collapsed `data:*;base64,...` and returned immediately, which skipped the existing token/userinfo/private-key passes for data URL header parameters.
P1: Fix: media payload collapse now feeds the collapsed header+marker through the common secret redaction pipeline. Added a regression for an `sk-*` token in the data URL header.
P2: Local lightweight validation: `opam exec -- ocamlformat --check lib/llm_provider/secret_redactor.ml` -> pass; `git diff --check` -> pass. CI after push is mostly green; `Build & Test (OCaml 5.4.1)` is still `IN_PROGRESS`.
P3: Mergeability: `MERGEABLE` / `BLOCKED` is draft/checks-pending; no conflict found.

[근거] `gh pr view 2444 --repo jeong-sik/oas --json headRefOid,isDraft,mergeable,mergeStateStatus` + `gh pr checks 2444 --repo jeong-sik/oas --json name,state,bucket,workflow,link,startedAt,completedAt`, 확인일시 2026-07-02T15:53:40Z, 신뢰도 High.

PR #2529, opened 2026-07-11
fix(streaming): stop retries after observable output

jeong-sik:
## Problem

`Complete.complete_stream_with_retry` reused the same `on_event` callback for every provider attempt. If an attempt emitted partial text, thinking, or a tool call and then returned a retryable transport error, the retry appended a second provider response to the same consumer-visible stream. This can produce repeated narration, duplicate tool starts, and orphan/missing tool-result projections regardless of model family.

## Change

- classify normalized SSE constructors by whether they commit observable assistant-turn state;
- retry only while an attempt has emitted no content, terminal signal, or typed stream failure;
- after the boundary, return the original typed `Http_client.http_error` and log the suppressed retry explicitly;
- preserve existing exponential retry behavior for pre-stream failures;
- document the transaction contract in `complete.mli`.

No provider/model name matching or retry-count heuristic is introduced.

## Verification

- provider-neutral event cases: text delta, thinking delta, tool start, tool argument delta;
- each case runs through Anthropic, OpenAI-compatible, and Gemini configurations;
- regression asserts one provider request, one observable event, and preservation of the original typed error;
- pre-event retry regression asserts the second attempt still succeeds;
- `ocamlformat --check` passed for all touched OCaml files;
- `git diff --check` passed.

Per repository policy, no local Dune build/test was run; CI is the build authority.

## Follow-up kept separate

ID-less streaming tool calls can expose `tool_id=None` at block start and a different synthesized ID at finalize. That is a separate identity-contract issue: #2528.

### DIFF
--- a/lib/llm_provider/complete.ml
+++ b/lib/llm_provider/complete.ml
@@ -346,6 +346,30 @@ let complete_stream
       (ensure_nonempty_completion result)
 ;;
 
+(* A streaming retry can reuse the caller's [on_event] callback only while the
+   failed attempt is still observationally empty.  Once content, a terminal
+   signal, or a typed stream failure has crossed that callback boundary, OAS has
+   no rollback event with which to retract it.  Retrying at that point would
+   splice two provider attempts into one consumer-visible stream.
+
+   Prelude-only events do not commit an attempt: [Connected], [Ping], and
+   [MessageStart] carry no assistant content and leave the content accumulator
+   empty.  Every other constructor either mutates the assistant turn, closes
+   part of it, or reports a terminal failure to the consumer. *)
+let stream_event_commits_attempt : Types.sse_event -> bool = function
+  | Types.Connected | Types.Ping | Types.MessageStart _ -> false
+  | Types.ContentBlockStart _
+  | Types.ContentBlockDelta _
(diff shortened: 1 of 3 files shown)


jeong-sik:
**[재검증 @ `374e658ff2`] P1·P2 둘 다 해소 — CI green 시 병합 가능**

이전 P1(MessageStart splice)이 `374e658ff2 "treat MessageStart as a commit boundary"`에서 정확히 닫혔어용:
- `stream_event_commits_attempt`에서 `MessageStart _`를 committing branch로 이동, `Connected|Ping`만 rollback-safe prelude로 남겼어요. mli도 갱신.
- `test_complete_retry.ml`에 MessageStart retry-suppression 테스트를 추가해서 P2(테스트 gap)도 동시에 닫혔어요.

MessageStart는 provider message identity(id/model/usage)를 consumer callback에 commit하니, 이제 재시도가 그 뒤로 splice되지 않아요. PR docstring 계약과 일치해요.

- P0/P1/P2/P3: 전부 없음(해소)
- 병합 가능성: **CI green이면 가능**해요. 현재 Build & Test(5.4.1/5.5.0)가 도는 중이에요.

PR #2820, opened 2026-07-27
feat(exact-output): own durable measurement receipts

jeong-sik:
## Why

MASC must durably journal OAS measurement dispatch intent and terminal evidence without duplicating the OAS receipt schema or reconstructing provider/model metadata.

## Contract

- adds the sole current-schema `Exact_output.measurement_receipt_snapshot` codec with SHA-256 integrity
- adds stable operation and immutable receipt accessors
- classifies dispatch intent, terminal advancement, idempotent replay, operation/binding conflicts, invalid commit phases, phase regression, and evidence conflict
- rejects missing, extra, malformed, future-version, internally inconsistent, and corrupted evidence

## Forward cut

- replaces the snapshot's full resolved `visit` with provider-neutral stable flow/ordinal/candidate-binding identity
- keeps type, codec, and transition ownership in Dune-private `Exact_output_measurement_receipt`; `Agent_sdk.Exact_output` is the sole public surface
- adds no compatibility decoder, default fill, migration, facade, provider/model branch, or Pricing logic

## Validation

- `scripts/dune-local.sh build test/test_exact_output_flow.exe`
- `scripts/dune-local.sh exec test/test_exact_output_flow.exe -- test outer-flow 27`
- `bash scripts/ci-code-smell-ratchet.sh --check`
- `ocamlformat --check lib/llm_provider/exact_output_measurement_receipt.mli lib/llm_provider/exact_output_measurement_receipt.ml lib/llm_provider/exact_output.mli lib/llm_provider/exact_output.ml test/test_exact_output_flow.ml`
- `git diff --check`

### DIFF
--- a/lib/llm_provider/exact_output.ml
+++ b/lib/llm_provider/exact_output.ml
@@ -21,7 +21,7 @@ type measurement_evidence = Flow_admission.measurement_evidence =
   ; outcome : measurement_outcome
   }
 
-type measurement_operation_id = Flow_admission.measurement_operation_id
+type measurement_operation_id = Measurement_operation_id of string
 
 type measurement_receipt_phase = Flow_admission.measurement_receipt_phase =
   | Measurement_fence_committed
@@ -160,12 +160,39 @@ type flow_measurement_receipt =
 
 type measurement_receipt_snapshot =
   { operation_id : measurement_operation_id
-  ; visit : flow_candidate_visit
+  ; flow_id : flow_id
+  ; visit_ordinal : flow_visit_ordinal
+  ; candidate_id : string
+  ; candidate_binding_sha256 : string
   ; request_body_sha256 : string
   ; phase : measurement_receipt_phase
   ; dispatch : measurement_dispatch_fact
   ; outcome : measurement_outcome option
   }
+
+type measurement_receipt_snapshot_decode_error =
+  | Measurement_receipt_snapshot_malformed_json of string
+  | Measurement_receipt_snapshot_invalid_fields
+  | Measurement_receipt_snapshot_unknown_format of string
+  | Measurement_receipt_snapshot_unsupported_version of int
+  | Measurement_receipt_snapshot_invalid_field of string
+  | Measurement_receipt_snapshot_integrity_mismatch
+
+type measurement_receipt_transition_conflict =
+  | Measurement_operation_mismatch
+  | Measurement_operation_binding_mismatch
+  | Measurement_invalid_commit_phase of measurement_receipt_phase
+  | Measurement_phase_regression of
+      { previous_phase : measurement_receipt_phase
+      ; incoming_phase : measurement_receipt_phase
+      }
+  | Measurement_evidence_conflict
+
+type measurement_receipt_transition =
+  | Measurement_dispatch_intent
+  | Measurement_terminal_advance
+  | Measurement_idempotent_replay
+  | Measurement_transition_conflict of measurement_receipt_transition_conflict
 
 type flow_candidate_step =
   { visit : flow_candidate_visit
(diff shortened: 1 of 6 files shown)


jeong-sik (commented):
Adversarial exact-flow review at current head `cf1e479387d075da935806e3f8e90c6dff71c373` found two blocking contract issues; see inline comments. The added test also does not exercise the PR-described missing/extra/malformed/internal-consistency rejection matrix or invalid previous durable boundaries, so those cases should be added with the fixes. No approval was submitted.

lib/llm_provider/exact_output_measurement_receipt.ml
```diff
@@ -0,0 +1,416 @@
...
+    then Measurement_transition_conflict Measurement_operation_binding_mismatch
+    else if same_snapshot previous incoming
+    then Measurement_idempotent_replay
+    else if phase_rank incoming.phase < phase_rank previous.phase
+    then
+      Measurement_transition_conflict
+        (Measurement_phase_regression
+           { previous_phase = previous.phase; incoming_phase = incoming.phase })
+    else if
+      previous.phase <> Measurement_terminal
```
jeong-sik: **[P1] Reject invalid previous durable boundaries before terminal advance.** Both `wire_started/dispatch_started/None` and `fence_committed/no_dispatch/None` pass `shape_is_valid` and the current-schema decoder. With either snapshot as `previous`, this broad non-terminal check returns `Measurement_terminal_advance`. That contradicts the public contract that intermediate live observations are not durable callback boundaries: a journal that accidentally committed one is silently legitimized instead of failing closed. Please gate terminal advancement on `is_dispatch_intent previous` and classify every other non-terminal previous snapshot as an invalid commit/evidence conflict; add cases for both shapes.

lib/llm_provider/exact_output.ml
```diff
@@ -479,32 +442,28 @@ let flow_success_output success = success.success
...
-let same_measurement
-      (left : measurement_receipt_snapshot)
-      (right : measurement_receipt_snapshot)
-  =
-  String.equal
-    (measurement_operation_id_to_string left.operation_id)
-    (measurement_operation_id_to_string right.operation_id)
-;;
+  let candidate =
+    Flow_contract.flow_preference_identity_of_candidate measurement.visit.identity
```
jeong-sik, line 452: **[P1] Preserve the resolved catalog provenance in the durable binding.** `flow_preference_identity_of_candidate` is intentionally a preference projection: its binding hash is only `target_identity_fingerprint`. The prior `visit` also carried `catalog_generation` and `catalog_evidence`, so flattening through this helper irreversibly removes the exact resolver generation/evidence from the v1 journal. Two catalog snapshots with the same target binding can no longer be attributed during an OAS audit. This does not require provider/model leakage: OAS already exposes provider-neutral `catalog_generation_fingerprint` and `catalog_evidence_sha256`. Please retain those values in the current schema/binding before the no-migration v1 is written.

lib/llm_provider/exact_output_measurement_receipt.ml
jeong-sik: Addressed in `0319f8b27`. `classify_measurement_receipt_transition` now accepts a previous durable boundary only when it is either a dispatch intent or a terminal receipt; decoded `wire_started/dispatch_started/None` and `fence_committed/no_dispatch/None` now fail closed as typed `Measurement_invalid_previous_boundary`. Focused coverage also pins wire-started commit rejection and conflicting terminal evidence. `outer-flow 27` passes.

lib/llm_provider/exact_output.ml
jeong-sik: Addressed in `0319f8b27`. Durable v1 now carries provider-neutral `catalog_generation_fingerprint` and `catalog_evidence_sha256`, validates both as SHA-256 evidence, includes both in binding equality, and populates them from the resolved catalog visit identity. Round-trip and binding-mismatch tests cover both fields; no provider/model identity was added.

PR #2915, opened 2026-08-01
fix(sse): keep first-event deadline until data arrives

jeong-sik:
## Summary

An SSE `event:` field selects the dispatch type but is metadata, not a payload-bearing event. The reader previously marked first output as seen on that line, switching from the caller-provided first-event/TTFT deadline to the shorter inter-token idle deadline before any `data:` field arrived.

This PR transitions to the inter-token idle budget only after a `data:` field and adds a deterministic mock-clock regression test.

## Boundary

- OAS-only change.
- No provider/model heuristic or error-string classification.
- No MASC dependency or lifecycle decision.
- Existing typed malformed SSE / NDJSON errors remain unchanged.

## Protocol reference

The [WHATWG Server-sent events specification](https://html.spec.whatwg.org/multipage/server-sent-events.html) treats `event:` as event-type metadata; dispatch requires accumulated data. The reader now keeps the first-event deadline until payload data arrives.

## Validation

- `test_sse_budget_anchor.exe`: 6/6
- `test_http_client.exe`: 34/34
- `test_complete_http.exe`: 67/67
- `git diff --check`: clean

Draft for CI and adversarial review.

### DIFF
--- a/lib/llm_provider/http_client.ml
+++ b/lib/llm_provider/http_client.ml
@@ -2602,14 +2602,15 @@ let read_sse
       | None, _ -> inner ()
     in
     (* P3a (RFC-OAS-037 review): the transition to the inter-token idle budget
-       must fire on GENUINE first output — a data/event field — NOT on a bare
-       blank dispatch delimiter. A provider that emits a leading blank line
-       before real prefill would otherwise switch to the short idle budget
-       prematurely and re-introduce the very bug this RFC fixes for it.
+       must fire on GENUINE first output — a data field — NOT on a bare event
+       type, blank dispatch delimiter, or other metadata. An [event] field
+       selects the dispatch type but carries no payload; counting it as the
+       first event would replace the caller's first-event bound with the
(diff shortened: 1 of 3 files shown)


jeong-sik (commented):
## 리뷰 — 커밋 `c7dac1f` (SSE first-event deadline until data)

### 방향: 동의 (SSE deadline 정확성, #2910/#2912 맥락)
SSE event field가 dispatch type을 선택하지만 payload-bearing이 아닌데, reader가 그 metadata line에서 first-event를 seen으로 표시 → TTFT deadline에서 inter-token idle deadline으로 **데이터 도착 전** 전환하던 버그. 이 PR은 data field 후에만 inter-token으로 전환한다.

- #2910(SSE joining)·#2912(EOF reuse)의 deadline 정확성 연장.
- 회귀 위험 낮음: 버그 수정(deadline 전환 시점 정확화). 테스트(`test_sse_budget_anchor`) 동반.

